PR #4, opened 2015-02-26
Run setuid as jetty with JETTY_BASE=/var/lib/jetty

### LINKED ISSUE #1: Run as non-root by default
Like most of the Docker official images, this image is currently configured to run as `root`. I'm opening this issue as a place to discuss what change should be made, if any. See [this discussion](https://github.com/docker-library/tomcat/issues/3) in the context of the `tomcat` image for the sort of issues that can be anticipated for `jetty`.

I can see a few changes/enhancements being made to this image:
1. Add code to the `Dockerfile` to create a consistent `jetty` user to allow users to easily do `USER jetty` in a derived `Dockerfile` or `-u jetty` in their `docker run` command.
2. Document the use of [`JETTY_BASE`](https://www.eclipse.org/jetty/documentation/current/startup-base-and-home.html) for the `jetty:9` image and give an example that uses `--module setuid` to drop privileges to the `jetty` user after starting as `root`.

md5:
This patch updates the `jetty` image to take advantage of the `JETTY_BASE` feature in Jetty 9.1+. Jetty now runs from `/var/lib/jetty`, which is owned by user `jetty`.

The `/var/lib/jetty` directory is created using Jetty's `start.jar --add-to-startd` functionality based on the list of modules in `$JETTY_HOME/start.ini`. The Dockerfile then adds Jetty's `setuid` to direct Jetty to drop privileges after startup.

Fixes #1

### DIFF
--- a/9-jre7/Dockerfile
+++ b/9-jre7/Dockerfile
@@ -24,5 +24,14 @@ RUN curl -SL "$JETTY_TGZ_URL" -o jetty.tar.gz \
     && rm -fr demo-base javadoc \
     && rm jetty.tar.gz*
 
+ENV JETTY_BASE /var/lib/jetty
+RUN mkdir -p "$JETTY_BASE" \
+    && chown jetty:jetty "$JETTY_BASE"
+WORKDIR $JETTY_BASE
+
+# Get the list of modules in the default start.ini and build new base with those modules, then add setuid
+RUN java -jar "$JETTY_HOME/start.jar" --add-to-startd="$(grep -- ^--module= "$JETTY_HOME/start.ini" | cut -d= -f2 | paste -d, -s)" \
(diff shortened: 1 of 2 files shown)


md5:
@tianon @yosifkit thoughts?

yosifkit:
That seems reasonable.

md5:
The image as built by this branch can actually be run directly with `-u jetty`; the `setuid` module just has no effect in that case.

The reason I didn't use `USER jetty` is probably the same reason that (for example) `ghost` uses `gosu` instead of `USER user`, i.e. ease of creating derived images without having to switch back to `USER root`. In this case, Jetty provides the functionality built-in, so I didn't want to add `gosu`.

I'm planning to document this stuff and submit PRs for both the image changes and the docs at the same time.

tianon:
Seems good to me too.  Apache and the mysql variations do a similar flow. :+1:

PR #5, opened 2015-02-28
Support running Jetty in read-only containers

md5:
This PR updates the `jetty` image to be able to easily run as a read-only container

```
docker run -d --read-only -v /tmp/jetty -v /run/jetty jetty
```

It makes the following changes:
1. Creates `/run/jetty` and `/tmp/jetty` directories owned by `jetty:jetty`
2. Sets `JETTY_STATE=/run/jetty/jetty.state` and `TMPDIR=/tmp/jetty`
3. <s>Declares `/run/jetty` and `/tmp/jetty` to be `VOLUME`s</s>

<s>The `VOLUME` declarations could be dropped and replaced with instructions to run with `--read-only -v /run/jetty -v /tmp/jetty`, but I'm in favor of these service-specific `/run` and `/tmp` directories being `VOLUME`s in general.</s>

Of course, the above command will run a pretty useless `jetty` container unless you also mount in `/var/lib/jetty/webapps` or otherwise configure the server to do something useful.

### DIFF
--- a/9-jre7/Dockerfile
+++ b/9-jre7/Dockerfile
@@ -25,8 +25,7 @@ RUN curl -SL "$JETTY_TGZ_URL" -o jetty.tar.gz \
     && rm jetty.tar.gz*
 
 ENV JETTY_BASE /var/lib/jetty
-RUN mkdir -p "$JETTY_BASE" \
-    && chown jetty:jetty "$JETTY_BASE"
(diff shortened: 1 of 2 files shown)


md5:
Making these volumes may not be the best idea since one or both of them should possibly be on `tmpfs`.

cf. https://github.com/docker/docker/pull/8478

PR #23, opened 2016-01-19
creating jetty container with java jdk7u79

agutoli:


md5:
@agutoli Did you mean to open this PR? If so, could you explain your motivation a bit more?

agutoli:
@md5 I did it just because I have a java environment with jdk7u79 approved, then I needed a local Dockerfile with this change into my project and thought that is a good opportunity to contribute with your project.

md5:
I'd like to keep this image using the current Java version, which for Java 7 is `7u91`. This version will update as the Docker Hub `java` image is updated.

Is your main issue needing to use `7u79` or is it needing to use the JDK image vs. JRE for some reason?

agutoli:
@md5 More because the JDK that we need to build a java project, the version  its ok if is 7u91 or 7u79...

md5:
@agutoli In that case, you should just be able to add the `openjdk-7-jdk` package on top of the existing `jetty:9.2-jre7` image as follows:

``` Dockerfile
FROM jetty:9.2-jre7
RUN apt-get update && apt-get install -y openjdk-7-jdk="$JAVA_DEBIAN_VERSION"
```

Since the `CMD` will happen after anything you do in a `RUN` directive, Jetty won't care if the JDK was installed after the JRE was already installed. Since `openjdk-7-jdk` has `openjdk-7-jre-headless` as a prerequisite (via `openjdk-7-jre`), the image size should be roughly the same either way.

PR #32, opened 2016-04-04
Add alpine-base images

hypnoce:
Use java alpine as base images for jetty.

### DIFF
--- a/9.3-jre8/alpine/Dockerfile
+++ b/9.3-jre8/alpine/Dockerfile
@@ -11,7 +11,7 @@ WORKDIR $JETTY_HOME
 ENV JETTY_BASE /var/lib/jetty
 RUN mkdir -p "$JETTY_BASE"
 
-ENV JETTY_VERSION 9.3.8.v20160314
+ENV JETTY_VERSION 9.3.9.v20160517
 ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-distribution/$JETTY_VERSION/jetty-distribution-$JETTY_VERSION.tar.gz
 
 # GPG Keys are personal keys of Jetty committers (see https://dev.eclipse.org/mhonarc/lists/jetty-users/msg05220.html)
--- a/9.3-jre8/alpine/Dockerfile
+++ b/9.3-jre8/alpine/Dockerfile
@@ -1,14 +1,17 @@
-FROM java:8-jre
+FROM java:8-jre-alpine
 
 # add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
-RUN groupadd -r jetty && useradd -r -g jetty jetty
+RUN addgroup -S jetty && adduser -D -S -H -G jetty jetty && rm -rf /etc/group- /etc/passwd- /etc/shadow-
 
 ENV JETTY_HOME /usr/local/jetty
 ENV PATH $JETTY_HOME/bin:$PATH
 RUN mkdir -p "$JETTY_HOME"
 WORKDIR $JETTY_HOME
 
-ENV JETTY_VERSION 9.2.15.v20160210
+ENV JETTY_BASE /var/lib/jetty
+RUN mkdir -p "$JETTY_BASE"
+
+ENV JETTY_VERSION 9.3.8.v20160314
 ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-distribution/$JETTY_VERSION/jetty-distribution-$JETTY_VERSION.tar.gz
 
 # GPG Keys are personal keys of Jetty committers (see https://dev.eclipse.org/mhonarc/lists/jetty-users/msg05220.html)
@@ -17,8 +20,11 @@ ENV JETTY_GPG_KEYS \
        B59B67FD7904984367F931800818D9D68FB67BAC \
        # 1024D/D7C58886 2010-03-09 Jesse McConnell (signing key) <jesse.mcconnell@gmail.com>
        5DE533CB43DAF8BC3E372283E7AE839CD7C58886
-
 RUN set -xe \
+
+	# Install required packages for build time. Will be removed when build finishes.
+	&& apk add --no-cache --virtual .build-deps gnupg coreutils curl \
+
 	&& curl -SL "$JETTY_TGZ_URL" -o jetty.tar.gz \
 	&& curl -SL "$JETTY_TGZ_URL.asc" -o jetty.tar.gz.asc \
         && export GNUPGHOME="$(mktemp -d)" \
@@ -26,19 +32,24 @@ RUN set -xe \
                 gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$key"; done \
 	&& gpg --batch --verify jetty.tar.gz.asc jetty.tar.gz \
 	&& rm -r "$GNUPGHOME" \
-	&& tar -xvf jetty.tar.gz --strip-components=1 \
+	&& tar -xvzf jetty.tar.gz \
+	&& mv jetty-distribution-$JETTY_VERSION/* ./ \
 	&& sed -i '/jetty-logging/d' etc/jetty.conf \
 	&& rm -fr demo-base javadoc \
-	&& rm jetty.tar.gz*
+	&& rm jetty.tar.gz* \
+	&& rm -fr jetty-distribution-$JETTY_VERSION/ \
 
-ENV JETTY_BASE /var/lib/jetty
-RUN mkdir -p "$JETTY_BASE"
+	# Get the list of modules in the default start.ini and build new base with those modules, then add setuid
+	&& cd $JETTY_BASE \
+	&& modules="$(grep -- ^--module= "$JETTY_HOME/start.ini" | cut -d= -f2 | paste -d, -s)" \
+	&& java -jar "$JETTY_HOME/start.jar" --add-to-startd="$modules,setuid" \
+
+	# Remove installed packages and various cleanup
+	&& apk del .build-deps \
+	&& rm -fr .build-deps \
+	&& rm -rf /tmp/hsperfdata_root
+
 WORKDIR $JETTY_BASE
-
-# Get the list of modules in the default start.ini and build new base with those modules, then add setuid
-RUN modules="$(grep -- ^--module= "$JETTY_HOME/start.ini" | cut -d= -f2 | paste -d, -s)" \
-	&& set -xe \
-	&& java -jar "$JETTY_HOME/start.jar" --add-to-startd="$modules,setuid"
 
 ENV JETTY_RUN /run/jetty
 ENV JETTY_STATE $JETTY_RUN/jetty.state
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -35,45 +35,54 @@ outputTag() {
 }
 
 for path in "${paths[@]}"; do
-	echo
+	for variant in '' alpine; do
+		echo
 
-	commit="$(git log -1 --format='format:%H' -- "$path")"
+		[ -f "$path${variant:+/$variant}/Dockerfile" ] || continue
 
-	suffix="${path#*-}" # "jre7"
+		commit="$(git log -1 --format='format:%H' -- "$path${variant:+/$variant}")"
 
-	version="$(grep -m1 'ENV JETTY_VERSION ' "$path/Dockerfile" | cut -d' ' -f3)"
+		suffix="${path#*-}" # "jre7"
 
-	if [[ "$version" == *.v* ]]; then
-		# Release version
-		versionAliases=()
-		while [[ "$version" == *.* ]]; do
-			version="${version%.*}"
-			versionAliases+=("$version")
-		done
-	else
-		# Non-release version
-		versionAliases=("$version")
-	fi
+		version="$(grep -m1 'ENV JETTY_VERSION ' "$path${variant:+/$variant}/Dockerfile" | cut -d' ' -f3)"
 
-	# Output ${versionAliases[@]} without suffixes
-	# e.g. 9.2.10, 9.2, 9
-	if [ "$suffix" = "$defaultSuffix" ]; then
+		if [[ "$version" == *.v* ]]; then
+			# Release version
+			versionAliases=()
+			while [[ "$version" == *.* ]]; do
+				version="${version%.*}"
+				versionAliases+=("$version")
+			done
+		else
+			# Non-release version
+			versionAliases=("$version")
+		fi
+
+		# Output ${versionAliases[@]} without suffixes
+		# e.g. 9.2.10, 9.2, 9, 9.3-alpine
+		if [ "$suffix" = "$defaultSuffix" ]; then
+			for va in "${versionAliases[@]}"; do
+				outputTag "$va${variant:+-$variant}" "$url" "$commit" "$path${variant:+/$variant}"
+			done
+		fi
+
+		# Output ${versionAliases[@]} with suffixes
+		# e.g. 9.2.10-jre7, 9.2-jre7, 9-jre7, 9-jre8-alpine
 		for va in "${versionAliases[@]}"; do
-			outputTag "$va" "$url" "$commit" "$path"
+			outputTag "$va-$suffix${variant:+-$variant}" "$url" "$commit" "$path${variant:+/$variant}"
 		done
-	fi
 
-	# Output ${versionAliases[@]} with suffixes
-	# e.g. 9.2.10-jre7, 9.2-jre7, 9-jre7
-	for va in "${versionAliases[@]}"; do
-		outputTag "$va-$suffix" "$url" "$commit" "$path"
+		# Output custom aliases
+		# e.g. latest, jre7, jre8, latest-alpine
+		if [ ${#aliases[$path]} -gt 0 ]; then
+			for va in ${aliases[$path]}; do
+				if [ ! -z "$variant" -a "$va" = 'latest' ]; then
+					va="$variant"
+				else
+					va="$va${variant:+-$variant}"
+				fi
+				outputTag "$va" "$url" "$commit" "$path${variant:+/$variant}"
+			done
+		fi
 	done
-
-	# Output custom aliases
-	# e.g. latest, jre7, jre8
-	if [ ${#aliases[$path]} -gt 0 ]; then
-		for va in ${aliases[$path]}; do
-			outputTag "$va" "$url" "$commit" "$path"
-		done
-	fi
 done
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,17 +4,18 @@ language: bash
 dist: trusty
 
 env:
-  - VERSION=9.2 VARIANT=jre7
-  - VERSION=9.2 VARIANT=jre8
-  - VERSION=9.3 VARIANT=jre8
+  - VERSION=9.2-jre7
+  - VERSION=9.2-jre8
+  - VERSION=9.3-jre8
+  - VERSION=9.3-jre8 VARIANT=alpine
 
 install:
   - git clone https://github.com/docker-library/official-images.git ~/official-images
 
 before_script:
   - env | sort
-  - cd "$VERSION-$VARIANT"
-  - image="jetty:$VERSION-$VARIANT"
+  - cd "${VERSION}${VARIANT:+/$VARIANT}"
+  - image="jetty:${VERSION}${VARIANT:+-$VARIANT}"
 
 script:
   - docker build --pull -t "$image" .
--- a/9.2-jre7/Dockerfile
+++ b/9.2-jre7/Dockerfile
@@ -47,8 +47,8 @@ RUN set -xe \
 	&& mkdir -p "$JETTY_RUN" "$TMPDIR" \
 	&& chown -R jetty:jetty "$JETTY_RUN" "$TMPDIR" "$JETTY_BASE"
 
-COPY docker-entrypoint.bash /
+COPY docker-entrypoint.sh /
 
 EXPOSE 8080
-ENTRYPOINT ["/docker-entrypoint.bash"]
+ENTRYPOINT ["/docker-entrypoint.sh"]
 CMD ["java","-Djava.io.tmpdir=/tmp/jetty","-jar","/usr/local/jetty/start.jar"]
--- a/9.2-jre7/docker-entrypoint.sh
+++ b/9.2-jre7/docker-entrypoint.sh
@@ -1,8 +1,16 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
 if [ "$1" = jetty.sh ]; then
+	if ! command -v bash >/dev/null 2>&1 ; then
+		cat >&2 <<- 'EOWARN'
+			********************************************************************
+			ERROR: bash not found. Use of jetty.sh requires bash.
+			********************************************************************
+		EOWARN
+		exit 1
+	fi
 	cat >&2 <<- 'EOWARN'
 		********************************************************************
 		WARNING: Use of jetty.sh from this image is deprecated and may
@@ -14,7 +22,7 @@ if [ "$1" = jetty.sh ]; then
 	EOWARN
 fi
 
-if ! type -- "$1" &>/dev/null; then
+if ! command -v -- "$1" >/dev/null 2>&1 ; then
 	set -- java -jar "-Djava.io.tmpdir=$TMPDIR" "$JETTY_HOME/start.jar" "$@"
 fi
 
--- a/9.2-jre8/docker-entrypoint.sh
+++ b/9.2-jre8/docker-entrypoint.sh
@@ -1,8 +1,16 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
 if [ "$1" = jetty.sh ]; then
+	if ! command -v bash >/dev/null 2>&1 ; then
+		cat >&2 <<- 'EOWARN'
+			********************************************************************
+			ERROR: bash not found. Use of jetty.sh requires bash.
+			********************************************************************
+		EOWARN
+		exit 1
+	fi
 	cat >&2 <<- 'EOWARN'
 		********************************************************************
 		WARNING: Use of jetty.sh from this image is deprecated and may
@@ -14,7 +22,7 @@ if [ "$1" = jetty.sh ]; then
 	EOWARN
 fi
 
-if ! type -- "$1" &>/dev/null; then
+if ! command -v -- "$1" >/dev/null 2>&1 ; then
 	set -- java -jar "-Djava.io.tmpdir=$TMPDIR" "$JETTY_HOME/start.jar" "$@"
 fi
 
--- a/9.3-jre8/Dockerfile
+++ b/9.3-jre8/Dockerfile
@@ -8,7 +8,7 @@ ENV PATH $JETTY_HOME/bin:$PATH
 RUN mkdir -p "$JETTY_HOME"
 WORKDIR $JETTY_HOME
 
-ENV JETTY_VERSION 9.3.8.v20160314
+ENV JETTY_VERSION 9.3.9.v20160517
 ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-distribution/$JETTY_VERSION/jetty-distribution-$JETTY_VERSION.tar.gz
 
 # GPG Keys are personal keys of Jetty committers (see https://dev.eclipse.org/mhonarc/lists/jetty-users/msg05220.html)
--- a/9.3-jre8/Dockerfile
+++ b/9.3-jre8/Dockerfile
@@ -46,8 +46,8 @@ RUN set -xe \
 	&& mkdir -p "$JETTY_RUN" "$TMPDIR" \
 	&& chown -R jetty:jetty "$JETTY_RUN" "$TMPDIR" "$JETTY_BASE"
 
-COPY docker-entrypoint.bash /
+COPY docker-entrypoint.sh /
 
 EXPOSE 8080
-ENTRYPOINT ["/docker-entrypoint.bash"]
+ENTRYPOINT ["/docker-entrypoint.sh"]
 CMD ["java","-Djava.io.tmpdir=/tmp/jetty","-jar","/usr/local/jetty/start.jar"]
--- a/9.3-jre8/alpine/docker-entrypoint.sh
+++ b/9.3-jre8/alpine/docker-entrypoint.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+
+set -e
+
+if [ "$1" = jetty.sh ]; then
+	if ! command -v bash >/dev/null 2>&1 ; then
+		cat >&2 <<- 'EOWARN'
+			********************************************************************
+			ERROR: bash not found. Use of jetty.sh requires bash.
+			********************************************************************
+		EOWARN
+		exit 1
+	fi
+	cat >&2 <<- 'EOWARN'
+		********************************************************************
+		WARNING: Use of jetty.sh from this image is deprecated and may
+			 be removed at some point in the future.
+
+			 See the documentation for guidance on extending this image:
+			 https://github.com/docker-library/docs/tree/master/jetty
+		********************************************************************
+	EOWARN
+fi
+
+if ! command -v -- "$1" >/dev/null 2>&1 ; then
+	set -- java -jar "-Djava.io.tmpdir=$TMPDIR" "$JETTY_HOME/start.jar" "$@"
+fi
+
+exec "$@"
--- a/9.3-jre8/docker-entrypoint.sh
+++ b/9.3-jre8/docker-entrypoint.sh
@@ -1,8 +1,16 @@
-#!/bin/bash
+#!/bin/sh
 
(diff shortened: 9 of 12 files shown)


md5:
Thanks for your contribution @hypnoce 

Would you mind looking at the test failures in Travis? Looks like the issue is that `bash` doesn't exist in the container.

md5:
Also, there will be changes needed for the `update.sh` and `generate-stackbrew-library.sh` scripts in order to move forward with the addition of Alpine images.

@gregw If you have any feelings one way or the other about having a `jetty` image based on `alpine`, please speak up! :+1:

gregw:
I think an alpine image is a great idea.  I've no experience with alpine, but the intention of it matches well with the philosophy of Jetty - no taxation without representation!

@hypnoce However, if you really want to get serious about the size, then basing the alpine jetty image on the standard distro of jetty is not the best that can be done.  It should be possible to trim unused jars out of the distro, or even to build up from a non-distro base, only adding the modules that are needed.   Jetty distro is almost 14MB unpacked, and while I've not done the trim for a while I'm confident that you can get a running server in < 2MB of jars.

hypnoce:
Hi,

I was trying to make tests pass, but it seems that on jetty 9.2.x, libsetuid-linux.so has a missing symbol : 
java.lang.UnsatisfiedLinkError: /usr/local/jetty/lib/setuid/libsetuid-linux.so: Error relocating /usr/local/jetty/lib/setuid/libsetuid-linux.so: __sprintf_chk: symbol not found

It works on 9.3.x version. I looked at https://github.com/eclipse/jetty.toolchain/blob/master/jetty-setuid/libsetuid-linux/pom.xml to check how it was compiled but nothing was changed from 9.2.x to 9.3.x. Is there any other compilation env that I'm not aware of ?

For now, I will remove alpine for 9.2.x.

Thanks

gregw:
That error looks to me as if the libsetuid-linux.so lib will not link with the libraries supplied with the java 7 used for the 9.2.x build?   So that may be the difference with 1.0.1 and 1.0.3, simply the compiler used.

I wonder why this isn't a problem for the ubuntu builds?

hypnoce:
@gregw It does not work with java8 either. Java version is not the issue here. It think it's alpine that does not ship with this symbol. Looking deeper into the so files here what I found : 

9.2.x version : 
T setJavaFieldString
U setrlimit@@GLIBC_2.2.5
U setuid@@GLIBC_2.2.5
**U __sprintf_chk@@GLIBC_2.3.4**
U __stack_chk_fail@@GLIBC_2.4
T throwNewJavaException

9.3.x version : 
T setJavaFieldString
U setrlimit@@GLIBC_2.2.5
U setuid@@GLIBC_2.2.5
U sprintf@@GLIBC_2.2.5
T throwNewJavaException

Seems like the GLIBC version has an impact on the symbols generated.

gregw:
If need be we can start jetty without setuid.

joakime:
Looks like we need to compile setuid on a system similar to Alpine.  The current build of 9.2.x seems to have been built on a  glibc environment too new for Alpine.

hypnoce:
Removing setuid support could work. But we would have to run jetty as a root user.
We could also install glibc on alpine. Seems like a workaround more than a proven solution.
Last solution would be to ensure compilation of libs does not create missing symbols for alpine.

Which one do you prefer ?

gregw:
Without setuid we can still switch users in the docker entry script.   The down side of this is that you can't open privileged ports, but that trends not to be done anyway and port mapping used instead. 

So the down side would be increased differences with the standard images.  @md5 what do you think of that? Or could we also remove setuid from the standard images?

md5:
Let me try to summarize the options mentioned so far, along with my feelings on each:
1. Change the distribution build of Jetty to be `musl`-compatible
2. Run the container as `root` (perhaps just the `alpine` variants)
3. Don't use `setuid` in the default configuration

The first option seems preferable to me, but I don't have the knowledge to understand what it entails. I believe this would have an ancillary benefit of making the stock Jetty distribution work with Alpine or Busybox in other scenarios.

I think we all agree that the second option is a non-starter.

The third option doesn't seem great to me, since it would involve changing the image to have `USER jetty`, which would complicate building derived images (and break any existing ones). I also feel the other benefits of using `setuid` are worthwhile, such as allowing the SSL key to be `root`-owned and not readable by the `jetty` user after startup.

In terms of other options, I'll propose a couple more:
1. Build a `musl`-compatible version of `setuid` in the `Dockerfile` itself
2. Allow `jetty` user to bind privileged ports by using `setcap` on the `java` binary

Of these options, I find the first to be simply a variant of the first option above, so overall it's my second choice. Again, I don't know the details of what this would mean practically, so perhaps it wouldn't work for some reason. Also, I'm not sure if there are other `*.so` files other than `setuid` that would need the same treatment for Alpine.

Regarding my second proposed option, I think it's a bit of a non-starter as well. I don't think anyone really cares about having Jetty in the container bind privileged ports. As I said before, I think the benefits of using `setuid` are more in the realm of having things in the container that the running server is not able to read like SSL keys. For background, there has been discussion about the privileged ports issue in https://github.com/docker/docker/issues/8460 and https://github.com/docker-library/logstash/pull/14. I believe part of the reason that hasn't been done is that some of the filesystem backends used by Docker don't support the capability bits that get set by `setcap`.

@joakime What are your thoughts on the difficulty of building an Alpine-compatible `setuid`, either to replace the one in the current Jetty distribution or a purpose-specific build just for the `alpine` variants of the `jetty` Docker image?

md5:
BTW, the reasoning behind my use of `setuid` can be found in #4, though @gregw and other Jetty maintainers were not involved in the image at that point. Here's the gist of the discussion:

> The image as built by this branch can actually be run directly with `-u jetty`; the `setuid` module just has no effect in that case.
> 
> The reason I didn't use `USER jetty` is probably the same reason that (for example) `ghost` uses `gosu` instead of `USER user`, i.e. ease of creating derived images without having to switch back to `USER root`. In this case, Jetty provides the functionality built-in, so I didn't want to add `gosu`.
> 
> I'm planning to document this stuff and submit PRs for both the image changes and the docs at the same time.

gregw:
@md5 I think we can build setuid versions for alpine.  Have to wait until @joakime is off the boat he currently is on... next week!

md5:
@gregw Sounds great :+1:

md5:
@gregw @joakime Any chance you guys have looked into building an Alpine-compatible version of `libsetuid-linux.so`?

gregw:
It is on the TODO list, but not this week I'm afraid.....   will get back
shortly with an ETA.

On 28 April 2016 at 08:07, Mike Dillon notifications@github.com wrote:

> @gregw https://github.com/gregw @joakime https://github.com/joakime
> Any chance you guys have looked into building an Alpine-compatible version
> of libsetuid-linux.so?
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> https://github.com/appropriate/docker-jetty/pull/32#issuecomment-215245807

## 

Greg Wilkins gregw@webtide.com CTO http://webtide.com

md5:
Thanks @gregw. Just wanted to check in.

jmcc0nn3ll:
Where can I find a docker container that this is supposed to run in to test out?

md5:
@jmcc0nn3ll You should be able to do something like this:

``` console
$ git clone -b alpine https://github.com/hypnoce/docker-jetty.git alpine-docker-jetty
$ cd alpine-docker-jetty
$ docker build -t jetty:9.3-alpine 9.3-jre8/alpine
```

I'd then probably mount in your local build of the `libsetuid-linux.so` file into the running image like so:

``` console
$ docker run -v $PWD/libsetuid-linux.so:/usr/local/jetty/lib/setuid/libsetuid-linux.so:ro jetty:9.3-alpine
```

md5:
@hypnoce I'd like to keep the entrypoint consistent across all the images since it gets copied by `update.sh`. Do you mind changing the top-level `docker-entrypoint.bash` to be a plain `*.sh` script as you've done in your `alpine` directory? It would also be great if you could modify `update.sh` to correctly update all Dockerfiles, including the one you're adding in this PR.

jmcc0nn3ll:
I am finally circling back around to this, resolved an issue in our setuid toolchain earlier and trying to get this linked up in the manner suggested.

$ docker run -v $PWD/target/lib:/usr/local/jetty/lib/setuid:ro jetty:9.3-alpine

This seems to be doing what is intended, running with a local directory mount of the contents in that directory in the container but I keep getting a CNFE on the jetty setuid listener goop.  I'll keep on it tomorrow and see what comes of it.

jmcc0nn3ll:
@md5 perhaps I am a bit confused on what is needed here...

If I just run the 'jetty:9.3-alpine' image things just work.  

> 2016-05-23 21:23:04.862:INFO::main: Logging initialized @233ms
> 2016-05-23 21:23:05.018:INFO:oejs.SetUIDListener:main: Setting umask=02
> 2016-05-23 21:23:05.024:INFO:oejs.SetUIDListener:main: Opened ServerConnector@340f438e{HTTP/1.1,[http/1.1]}{0.0.0.0:8080}
> 2016-05-23 21:23:05.024:INFO:oejs.SetUIDListener:main: Setting GID=101
> 2016-05-23 21:23:05.025:INFO:oejs.SetUIDListener:main: Setting UID=100
> 2016-05-23 21:23:05.027:INFO:oejs.Server:main: jetty-9.3.8.v20160314

Supporting 9.2.x is not particularly interesting on my book, it is based on java 7 which is EOL and 9.2.x itself is EOL from an open source perspective.  We support it for a handful of clients still using java 7 but even those are quickly updating to java 8 and jetty 9.3.x releases.

md5:
@jmcc0nn3ll Sounds reasonable. @gregw Does supporting an `alpine` image only for 9.3 sound fine to you?

hypnoce:
@md5 To keep docker-entrypoint consistent, do you mind removing the support for jetty.sh for all images since the alpine one does not support it ? 

```
if [ "$1" = jetty.sh ]; then
    cat >&2 <<- 'EOWARN'
        WARNING: Use of jetty.sh from this image is not supported.

             See the documentation for guidance on extending this image:
             https://github.com/docker-library/docs/tree/master/jetty
    EOWARN
    exit 1
fi
```

Another solution would be to change jetty.sh to use sh instead of bash.

gregw:
@md5 yes 9.3 only is fine. As @jmcc0nn3ll says 9.2 is EOL anyway.

gregw:
@hypnoce I don't think it is doable to make jetty.sh use bash.  The script is pretty dependent on the common distributions and is unlikely to work in alpine even if stripped.

The direct execution of `java -jar start.jar` is the preferred way to start the server and is what is used by the other images.  I don't really have a feel for how long we need to support the legacy jetty.sh approach and will defer to @md5 about that.

md5:
Looks like the transition away from `jetty.sh` happened in August 2015 (see #17).

Since we have no mechanism for informing users of a breaking change, I'm not sure when is a safe time to actually remove the `jetty.sh` logic from the entrypoint. If it isn't removed, I could see adding a check to the entrypoint to see if `bash` is installed and having it `exit 1` if `bash` is not installed and someone tries to run `jetty.sh`. I'm not sure how important that is, but it's also pretty harmless and would allow the entrypoint to be the same across image variants.

hypnoce:
@md5 Agree. Just made the last changes. 
I needed to keep curl installed on the alpine image to make the jetty-hello-web test pass. It adds 1MB to the image size.
We can use an alternative method using wget 

```
wget "http://jetty:8080/$url" --method "$method" -q -O /dev/stdout "@"
```

We can also add an option in the Dockerfile to remove curl, but this will mess it up. Or we can just keep it as it is.

md5:
@hypnoce This is looking great! Thanks for your efforts.

Regarding the `curl` issue, I think the right thing to do is to update `jetty-hello-web` to use a separate image for the `curl` client call. Originally, that test used an image called `appropriate/curl`, but we switched it to use the `jetty` image itself when we realized that it already had `curl`. I think the easiest fix would be to change the test to use `buildpack-deps:curl`. We may as well keep this image as small as possible. The fix for the test is a one-liner: https://github.com/docker-library/official-images/blob/master/test/tests/jetty-hello-web/run.sh#L11-L12

Here are the remaining issues I'd like to see fixed before we merge:
1. [x] Remove curl and fix the `jetty-hello-web` test to stop assuming all `jetty` images will have `curl`
2. [ ] Update to Jetty 9.3.9
3. [ ] Use tabs for all indentation to be consistent with the existing files
4. [ ] Squash all the changes into a single commit (or maybe one commit for the entrypoint changes and one for the `alpine` changes)

md5:
@gregw Regarding the removal of the warning for `jetty.sh`, what would you think about removing `jetty.sh` from the `jetty-distribution` package itself (perhaps in Jetty 9.4)? I don't know what the build pipeline for downstream `jetty` OS packages looks like, but if `jetty.sh` is not intended for any use other than being called from OS init scripts, it may not make sense to be in the distribution package at all.

jmcc0nn3ll:
We are kicking around some options like this.  One thought is to add a jetty distribution variant that is slimmed down and optimized more for this style of usage, without the demo-base, any documentation, etc.  Not having the jetty.sh in that new distro might make sense...or perhaps a streamlined script that doesn't have the historical OS bits and fobs of support.

md5:
@jmcc0nn3ll That sounds great :+1:

hypnoce:
@md5 Just made a [PR1780](https://github.com/docker-library/official-images/pull/1780) to use an image with curl.

md5:
Thanks @hypnoce 👍 

One additional task I remembered is that the `generate-stackbrew-library.sh` script will also need to be updated to include the `alpine` variant(s). If you'd like to do this as well, it will make it easier for me to submit the official images update. Otherwise it may take me a couple days to update the script myself before I can submit the request to the `docker-library` folks once we merge this.

hypnoce:
@md5 working on it !

md5:
I just did one more sanity check on the new image to make sure there weren't any unnecessary files creeping in.

Do you think we can prune the `/etc/group-`, `/etc/passwd-`, and `/etc/shadow-` files in the `addgroup`/`adduser` line? This line is also creating `/home/jetty`, which I'm not too worried about, though we could tell `adduser` to use `$JETTY_BASE` as the home directory instead. There's also the `/tmp/hsperfdata_root` directory being created by the call to `java -jar "$JETTY_HOME/start.jar" --add-to-startd=...`. It might be good to prune that.

None of these are a big deal in terms of image size, so I'd be fine overlooking them.

``` console
$ diff -u <(docker run --rm java:8-jre-alpine find / | sort) <(docker run --rm jetty:9.3-jre8-alpine find / | sort)
--- /dev/fd/63  2016-05-26 09:01:58.000000000 -0700
+++ /dev/fd/62  2016-05-26 09:01:58.000000000 -0700
@@ -98,6 +98,7 @@
 /dev/tty
 /dev/urandom
 /dev/zero
+/docker-entrypoint.sh
 /etc
 /etc/TZ
 /etc/X11
@@ -127,6 +128,7 @@
 /etc/crontabs/root
 /etc/fstab
 /etc/group
+/etc/group-
 /etc/hostname
 /etc/hosts
 /etc/init.d
@@ -153,6 +155,7 @@
 /etc/network/if-up.d
 /etc/os-release
 /etc/passwd
+/etc/passwd-
 /etc/periodic
 /etc/periodic/15min
 /etc/periodic/daily
@@ -169,6 +172,7 @@
 /etc/securetty
 /etc/services
 /etc/shadow
+/etc/shadow-
 /etc/shells
 /etc/ssl
 /etc/ssl/certs
@@ -534,6 +538,7 @@
 /etc/sysctl.d/00-alpine.conf
 /etc/udhcpd.conf
 /home
+/home/jetty
 /lib
 /lib/apk
 /lib/apk/db
@@ -1778,6 +1783,7 @@
 /proc/zoneinfo
 /root
 /run
+/run/jetty
 /sbin
 /sbin/acpid
 /sbin/adjtimex
@@ -13594,6 +13600,8 @@
 /sys/power/state
 /sys/power/wakeup_count
 /tmp
+/tmp/hsperfdata_root
+/tmp/jetty
 /usr
 /usr/bin
 /usr/bin/[
@@ -13991,6 +13999,221 @@
 /usr/local
 /usr/local/bin
 /usr/local/bin/docker-java-home
+/usr/local/jetty
+/usr/local/jetty/README.TXT
+/usr/local/jetty/VERSION.txt
+/usr/local/jetty/bin
+/usr/local/jetty/bin/jetty.sh
+/usr/local/jetty/etc
+/usr/local/jetty/etc/README.spnego
+/usr/local/jetty/etc/example-quickstart.xml
+/usr/local/jetty/etc/hawtio.xml
+/usr/local/jetty/etc/home-base-warning.xml
+/usr/local/jetty/etc/jamon.xml
+/usr/local/jetty/etc/jdbcRealm.properties
+/usr/local/jetty/etc/jetty-alpn.xml
+/usr/local/jetty/etc/jetty-annotations.xml
+/usr/local/jetty/etc/jetty-cdi.xml
+/usr/local/jetty/etc/jetty-debug.xml
+/usr/local/jetty/etc/jetty-debuglog.xml
+/usr/local/jetty/etc/jetty-deploy.xml
+/usr/local/jetty/etc/jetty-gcloud-sessions.xml
+/usr/local/jetty/etc/jetty-gzip.xml
+/usr/local/jetty/etc/jetty-http-forwarded.xml
+/usr/local/jetty/etc/jetty-http.xml
+/usr/local/jetty/etc/jetty-http2.xml
+/usr/local/jetty/etc/jetty-http2c.xml
+/usr/local/jetty/etc/jetty-https.xml
+/usr/local/jetty/etc/jetty-infinispan.xml
+/usr/local/jetty/etc/jetty-ipaccess.xml
+/usr/local/jetty/etc/jetty-jaas.xml
+/usr/local/jetty/etc/jetty-jdbc-sessions.xml
+/usr/local/jetty/etc/jetty-jmx-remote.xml
+/usr/local/jetty/etc/jetty-jmx.xml
+/usr/local/jetty/etc/jetty-logging.xml
+/usr/local/jetty/etc/jetty-lowresources.xml
+/usr/local/jetty/etc/jetty-monitor.xml
+/usr/local/jetty/etc/jetty-nosql.xml
+/usr/local/jetty/etc/jetty-plus.xml
+/usr/local/jetty/etc/jetty-proxy-protocol-ssl.xml
+/usr/local/jetty/etc/jetty-proxy-protocol.xml
+/usr/local/jetty/etc/jetty-proxy.xml
+/usr/local/jetty/etc/jetty-requestlog.xml
+/usr/local/jetty/etc/jetty-rewrite.xml
+/usr/local/jetty/etc/jetty-setuid.xml
+/usr/local/jetty/etc/jetty-spring.xml
+/usr/local/jetty/etc/jetty-ssl-context.xml
+/usr/local/jetty/etc/jetty-ssl.xml
+/usr/local/jetty/etc/jetty-started.xml
+/usr/local/jetty/etc/jetty-stats.xml
+/usr/local/jetty/etc/jetty.conf
+/usr/local/jetty/etc/jetty.xml
+/usr/local/jetty/etc/jminix.xml
+/usr/local/jetty/etc/jolokia.xml
+/usr/local/jetty/etc/krb5.ini
+/usr/local/jetty/etc/spnego.conf
+/usr/local/jetty/etc/spnego.properties
+/usr/local/jetty/etc/webdefault.xml
+/usr/local/jetty/lib
+/usr/local/jetty/lib/annotations
+/usr/local/jetty/lib/annotations/asm-5.0.1.jar
+/usr/local/jetty/lib/annotations/asm-commons-5.0.1.jar
+/usr/local/jetty/lib/annotations/javax.annotation-api-1.2.jar
+/usr/local/jetty/lib/apache-jsp
+/usr/local/jetty/lib/apache-jsp/org.eclipse.jdt.core.compiler.ecj-4.4.2.jar
+/usr/local/jetty/lib/apache-jsp/org.eclipse.jetty.apache-jsp-9.3.8.v20160314.jar
+/usr/local/jetty/lib/apache-jsp/org.mortbay.jasper.apache-el-8.0.27.jar
+/usr/local/jetty/lib/apache-jsp/org.mortbay.jasper.apache-jsp-8.0.27.jar
+/usr/local/jetty/lib/apache-jstl
+/usr/local/jetty/lib/apache-jstl/org.apache.taglibs.taglibs-standard-impl-1.2.5.jar
+/usr/local/jetty/lib/apache-jstl/org.apache.taglibs.taglibs-standard-spec-1.2.5.jar
+/usr/local/jetty/lib/cdi-core-9.3.8.v20160314.jar
+/usr/local/jetty/lib/cdi-servlet-9.3.8.v20160314.jar
+/usr/local/jetty/lib/cdi-websocket-9.3.8.v20160314.jar
+/usr/local/jetty/lib/ext
+/usr/local/jetty/lib/ext/.donotdelete
+/usr/local/jetty/lib/fcgi
+/usr/local/jetty/lib/fcgi/fcgi-client-9.3.8.v20160314.jar
+/usr/local/jetty/lib/fcgi/fcgi-server-9.3.8.v20160314.jar
+/usr/local/jetty/lib/http2
+/usr/local/jetty/lib/http2/http2-common-9.3.8.v20160314.jar
+/usr/local/jetty/lib/http2/http2-hpack-9.3.8.v20160314.jar
+/usr/local/jetty/lib/http2/http2-server-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jaspi
+/usr/local/jetty/lib/jaspi/javax.security.auth.message-1.0.0.v201108011116.jar
+/usr/local/jetty/lib/jetty-alpn-server-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-annotations-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-client-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-continuation-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-deploy-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-gcloud-session-manager-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-http-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-infinispan-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-io-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-jaas-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-jaspi-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-jmx-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-jndi-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-nosql-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-plus-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-proxy-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-quickstart-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-rewrite-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-schemas-3.1.jar
+/usr/local/jetty/lib/jetty-security-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-server-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-servlet-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-servlets-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-util-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-webapp-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jetty-xml-9.3.8.v20160314.jar
+/usr/local/jetty/lib/jndi
+/usr/local/jetty/lib/jndi/javax.mail.glassfish-1.4.1.v201005082020.jar
+/usr/local/jetty/lib/jndi/javax.transaction-api-1.2.jar
+/usr/local/jetty/lib/monitor
+/usr/local/jetty/lib/monitor/jetty-monitor-9.3.8.v20160314.jar
+/usr/local/jetty/lib/servlet-api-3.1.jar
+/usr/local/jetty/lib/setuid
+/usr/local/jetty/lib/setuid/jetty-setuid-java-1.0.3.jar
+/usr/local/jetty/lib/setuid/libsetuid-linux.so
+/usr/local/jetty/lib/setuid/libsetuid-osx.so
+/usr/local/jetty/lib/spring
+/usr/local/jetty/lib/spring/jetty-spring-9.3.8.v20160314.jar
+/usr/local/jetty/lib/websocket
+/usr/local/jetty/lib/websocket/javax-websocket-client-impl-9.3.8.v20160314.jar
+/usr/local/jetty/lib/websocket/javax-websocket-server-impl-9.3.8.v20160314.jar
+/usr/local/jetty/lib/websocket/javax.websocket-api-1.0.jar
+/usr/local/jetty/lib/websocket/websocket-api-9.3.8.v20160314.jar
+/usr/local/jetty/lib/websocket/websocket-client-9.3.8.v20160314.jar
+/usr/local/jetty/lib/websocket/websocket-common-9.3.8.v20160314.jar
+/usr/local/jetty/lib/websocket/websocket-server-9.3.8.v20160314.jar
+/usr/local/jetty/lib/websocket/websocket-servlet-9.3.8.v20160314.jar
+/usr/local/jetty/license-eplv10-aslv20.html
+/usr/local/jetty/logs
+/usr/local/jetty/logs/.donotdelete
+/usr/local/jetty/modules
+/usr/local/jetty/modules/.donotdelete
+/usr/local/jetty/modules/alpn-impl
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_05.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_11.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_20.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_25.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_31.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_40.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_45.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_51.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_60.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_65.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_66.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_71.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_72.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_73.mod
+/usr/local/jetty/modules/alpn-impl/alpn-1.8.0_74.mod
+/usr/local/jetty/modules/alpn.mod
+/usr/local/jetty/modules/annotations.mod
+/usr/local/jetty/modules/apache-jsp.mod
+/usr/local/jetty/modules/apache-jstl.mod
+/usr/local/jetty/modules/cdi.mod
+/usr/local/jetty/modules/client.mod
+/usr/local/jetty/modules/continuation.mod
+/usr/local/jetty/modules/debug.mod
+/usr/local/jetty/modules/debuglog.mod
+/usr/local/jetty/modules/deploy.mod
+/usr/local/jetty/modules/ext.mod
+/usr/local/jetty/modules/fcgi.mod
+/usr/local/jetty/modules/gcloud-sessions.mod
+/usr/local/jetty/modules/gzip.mod
+/usr/local/jetty/modules/hawtio.mod
+/usr/local/jetty/modules/home-base-warning.mod
+/usr/local/jetty/modules/http-forwarded.mod
+/usr/local/jetty/modules/http.mod
+/usr/local/jetty/modules/http2.mod
+/usr/local/jetty/modules/http2c.mod
+/usr/local/jetty/modules/https.mod
+/usr/local/jetty/modules/infinispan.mod
+/usr/local/jetty/modules/ipaccess.mod
+/usr/local/jetty/modules/jaas.mod
+/usr/local/jetty/modules/jamon.mod
+/usr/local/jetty/modules/jaspi.mod
+/usr/local/jetty/modules/jdbc-sessions.mod
+/usr/local/jetty/modules/jminix.mod
+/usr/local/jetty/modules/jmx-remote.mod
+/usr/local/jetty/modules/jmx.mod
+/usr/local/jetty/modules/jndi.mod
+/usr/local/jetty/modules/jolokia.mod
+/usr/local/jetty/modules/jsp.mod
+/usr/local/jetty/modules/jstl.mod
+/usr/local/jetty/modules/jvm.mod
+/usr/local/jetty/modules/logging.mod
+/usr/local/jetty/modules/lowresources.mod
+/usr/local/jetty/modules/monitor.mod
+/usr/local/jetty/modules/nosql.mod
+/usr/local/jetty/modules/plus.mod
+/usr/local/jetty/modules/proxy-protocol-ssl.mod
+/usr/local/jetty/modules/proxy-protocol.mod
+/usr/local/jetty/modules/proxy.mod
+/usr/local/jetty/modules/quickstart.mod
+/usr/local/jetty/modules/requestlog.mod
+/usr/local/jetty/modules/resources.mod
+/usr/local/jetty/modules/rewrite.mod
+/usr/local/jetty/modules/security.mod
+/usr/local/jetty/modules/server.mod
+/usr/local/jetty/modules/servlet.mod
+/usr/local/jetty/modules/servlets.mod
+/usr/local/jetty/modules/setuid.mod
+/usr/local/jetty/modules/spring.mod
+/usr/local/jetty/modules/ssl.mod
+/usr/local/jetty/modules/stats.mod
+/usr/local/jetty/modules/webapp.mod
+/usr/local/jetty/modules/websocket.mod
+/usr/local/jetty/notice.html
+/usr/local/jetty/resources
+/usr/local/jetty/resources/log4j.properties
+/usr/local/jetty/start.ini
+/usr/local/jetty/start.jar
+/usr/local/jetty/webapps
+/usr/local/jetty/webapps/README.TXT
 /usr/local/lib
 /usr/local/share
 /usr/local/share/ca-certificates
@@ -14602,6 +14825,22 @@
 /var/empty
 /var/lib
 /var/lib/apk
+/var/lib/jetty
+/var/lib/jetty/lib
+/var/lib/jetty/lib/ext
+/var/lib/jetty/resources
+/var/lib/jetty/start.d
+/var/lib/jetty/start.d/deploy.ini
+/var/lib/jetty/start.d/ext.ini
+/var/lib/jetty/start.d/home-base-warning.ini
+/var/lib/jetty/start.d/http.ini
+/var/lib/jetty/start.d/jsp.ini
+/var/lib/jetty/start.d/jstl.ini
+/var/lib/jetty/start.d/resources.ini
+/var/lib/jetty/start.d/server.ini
+/var/lib/jetty/start.d/setuid.ini
+/var/lib/jetty/start.d/websocket.ini
+/var/lib/jetty/webapps
 /var/lib/misc
 /var/lib/udhcpd
 /var/lock
```

hypnoce:
@md5 Do you want a jetty:alpine image or do you prefer to keep jetty:latest-alpine ?

md5:
@hypnoce I think we want the following for parity with the existing tags:
- `9.3.9-alpine`
- `9.3-alpine`
- `9-alpine`
- `9.3.9-jre8-alpine`
- `9.3-jre8-alpine`
- `9-jre8-alpine`
- `alpine`
- `jre8-alpine`

I'm not aware of any other official images using `latest-alpine`; they all use `alpine` for that purpose from what I've seen.

9.3-jre8/alpine/Dockerfile
```diff
@@ -44,9 +44,10 @@ RUN set -xe \
 	&& modules="$(grep -- ^--module= "$JETTY_HOME/start.ini" | cut -d= -f2 | paste -d, -s)" \
 	&& java -jar "$JETTY_HOME/start.jar" --add-to-startd="$modules,setuid" \
 
-	# Remove installed packages
+	# Remove installed packages and various cleanup
 	&& apk del .build-deps \
-	&& rm -fr .build-deps
+	&& rm -fr .build-deps \
+	&& rm -rf /tmp/hsperfdata_root /etc/group- /etc/passwd- /etc/shadow-
```
md5: Deleting the `/etc/group-` and other related files in this layer isn't ideal. They need to be deleted in the same layer they're added, which is the `adduser` layer above 👍

generate-stackbrew-library.sh
```diff
@@ -35,45 +35,49 @@ outputTag() {
 }
 
 for path in "${paths[@]}"; do
-	echo
+	for variant in alpine ''; do
```
md5: Can we switch this to list the non-Alpine images first? This file controls the order they show up in the generated docs on Docker Hub.

md5:
https://github.com/docker-library/official-images/issues/1780 has been merged, so I re-ran the Travis build and it now passes.

hypnoce:
@md5 All tasks completed.
Regarding the tab vs space, some files were already using mixed indentation. I left them as is. I just made sure the line I added were using tabs.
If you want, I can also remove the other space indentations.

md5:
Thanks @hypnoce!

md5:
I've opened https://github.com/docker-library/official-images/pull/1782 to get this pushed to Docker Hub :metal:

md5:
And it's live!

```
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
jetty               alpine              3610a079a378        31 minutes ago      117.4 MB
jetty               latest              a5c940ddabf2        39 minutes ago      320.4 MB
```

PR #41, opened 2016-06-29
Add pax flags to java binary

el-chogo:
Add pax flags so systems with grsec patchset may run jetty without any further modification (hopefully).

### DIFF
--- a/9.3-jre8/alpine/Dockerfile
+++ b/9.3-jre8/alpine/Dockerfile
@@ -23,7 +23,7 @@ ENV JETTY_GPG_KEYS \
 
 RUN set -xe \
 	# Install required packages for build time. Will be removed when build finishes.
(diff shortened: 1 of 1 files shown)


md5:
@clinoge Thanks for bringing up this issue, but I wonder if it would make more sense for these flags to be set in the `java:*-alpine` images instead of here.

Also, whether those flags are set in this image or in `java`, I'm not sure that we can rely on `setfattr` in official images yet, since my understanding is that AUFS didn't support extended attributes until quite recently.

If this approach works for your use case, it seems like you could easily call `setfattr` in an image derived from the `jetty` image. Others who need PaX support could do the same in their images (hopefully after seeing your approach documented in this PR).

el-chogo:
Wow, that's actually true, I should've put that on java:*-alpine. Must've been really sleepy ;).

md5:
Closing this request since it needs to happen in the `java` image if it can be done at all.

PR #43, opened 2016-07-01
Fix #37 Allow JVM options to be set via JAVA_OPTIONS

### LINKED ISSUE #37: How to configure Java heap
Is there a way to configure Java heap, or how to pass custom flags to the java VM?

gregw:
Fixes #37 by allowing JAVA_OPTIONS to be set.
Since this takes the approach of the entry point script modifying all java commands, this also moves the setting of TMPDIR to a conditional adjustment of JAVA_OPTIONS.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -23,7 +23,19 @@ if [ "$1" = jetty.sh ]; then
 fi
 
 if ! command -v -- "$1" >/dev/null 2>&1 ; then
-	set -- java "-Djava.io.tmpdir=$TMPDIR" -jar "$JETTY_HOME/start.jar" "$@"
+	set -- java -jar "$JETTY_HOME/start.jar" "$@"
+fi
+
+if [ -n "$TMPDIR" ] ; then
+	case "$JAVA_OPTIONS" in
+		*-Djava.io.tmpdir=*) ;;
+		*) JAVA_OPTIONS="-Djava.io.tmpdir=$TMPDIR $JAVA_OPTIONS" ;;
+	esac
+fi
+
+if [ "$1" = "java" -a -n "$JAVA_OPTIONS" ] ; then
+        shift
+	set -- java $JAVA_OPTIONS "$@"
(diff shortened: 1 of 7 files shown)


docker-entrypoint.sh
```diff
@@ -23,7 +23,19 @@ if [ "$1" = jetty.sh ]; then
 fi
 
 if ! command -v -- "$1" >/dev/null 2>&1 ; then
-	set -- java "-Djava.io.tmpdir=$TMPDIR" -jar "$JETTY_HOME/start.jar" "$@"
+	set -- java -jar "$JETTY_HOME/start.jar" "$@"
+fi
+
+if [ -n "$TMPDIR" ] ; then
```
md5, line 29: Since `TMPDIR` is set in the Dockerfile, this is basically always going to be true unless someone passes `-e TMPDIR=`, which seems quite unlikely.

In addition to that observation, I also feel like we may as well clean up some of the `jetty.sh`-related stuff in the `Dockerfile` now. `JETTY_STATE` no longer has any effect, since we are no longer passing `jetty.state` to the `start.jar` command. `JETTY_RUN` also has no effect, since all it was being used for previously was the `JETTY_PID` functionality in `jetty.sh`.

md5:
These changes look good, regardless of whether you want to amend the handling of `TMPDIR` to always assume its `/tmp/jetty`.

The only thing I'd ask before you merge is to squash the fixup commits into the first one.

md5:
Once you've merged, feel free to open a PR against `docker-library/official-images` to get the image updated. The `generate-stackbrew-library.sh` script assumes a GNU system, so you can run it directly under a GNU/Linux system. I'm on OS X, so I run `make library` to avoid having to hack a bunch of GNU-on-Mac garbage into the script.

docker-entrypoint.sh
gregw, line 29: If the image is extended, the TMPDIR may be cleared and the dockerfile reused.

PR #44, opened 2016-07-06
Issue 37 Allow JVM options to be set via JAVA_OPTIONS

gregw:
Updated the alpine image with same changes previously made for #37
removed the unused ENV variables JETTY_RUN and JETTY_STATE


md5:
@gregw Would you mind rebasing this?

gregw:
@md5 will try!  I thought the merge should have done that, but my git-foo is not the best.   Let's see what I can mess up with a rebase:)  Stand-by

gregw:
@md5 nope - that's still ugly.   Let me do this from a fresh branch so we only see the new commit

PR #45, opened 2016-07-07
Fix #37 Allow JVM options to be set via JAVA_OPTIONS

### LINKED ISSUE #37: How to configure Java heap

gregw:
Updated the alpine image with same changes previously made for #37
removed the unused ENV variables JETTY_RUN and JETTY_STATE


gregw:
@md5 ah this is much better!  Only a single commit.

md5:
@gregw Looks good to merge 👍

PR #50, opened 2016-09-30
Update to Jetty 9.2.19 and 9.3.12

md5:


md5:
See https://github.com/docker-library/official-images/pull/2214

Issue #34, opened 2016-06-10
Supported Jetty versions


md5:
I just noticed that there are 9.4.0.M0 and 9.3.10.M0 versions of Jetty after seeing [a comment](https://github.com/appropriate/docker-jetty/issues/30#issuecomment-225283574) from @joakime on #30. I also recall @jmcc0nn3ll commenting that Jetty 9.2 is considered EOL (see [here](https://github.com/appropriate/docker-jetty/pull/32#issuecomment-221319444)).

I wanted to open this issue in anticipation of the update to Jetty 9.4 to see which versions we should actually support.

joakime:
any Jetty release with `#.#.#.M#` is a milestone / unstable / beta release.

Jetty 9.4 will introduce a few things: easier configuration for embedded users (implemented), javax.websocket 1.1 support (being worked on now), websocket proxy support (planned), including support for JDK9 and it's new JVM level ALPN (implemented).

Jetty 10.x (also underway in master branch) will be Servlet 4.x

jmcc0nn3ll:
I also have the striped down jetty-home distribution almost ready to go, probably M1 which we can use in Alpine to shrink the size a bit.

md5:
@joakime Thanks for the explanation about the `M*` releases. I know that they are milestone releases and I don't generally build docker images for them, although I did build one for 9.3.0.M2. When those images _are_ built, they are only tagged with their full version number, so if we were to build a 9.4 image before the official release, it would not be tagged `jetty:9.4`, but only `jetty:9.4.0.M#`.

@jmcc0nn3ll That's great to hear! That may actually be a good reason to start early on a milestone-based `jetty:9.4.0.M#` image once those changes land in a milestone, since it will greatly simplify the `Dockerfile`.

md5:
As for a `jre9` image to complement the Java 9 support in Jetty 9.4, that's going to depend on https://github.com/docker-library/openjdk/issues/29. I haven't checked up on the status of OpenJDK 9 packaging in Debian for a while.

jmcc0nn3ll:
sounds good, I'll help get it wired up when we push it out, maybe a week or
two

## 

jesse mcconnell
jesse.mcconnell@gmail.com

md5:
[...]

PR #51, opened 2016-12-10
Add Jetty 9.4 images

### LINKED ISSUE #52: Support Jetty 9.4
Currently there's no tag for version 9.4 of Jetty.

hypnoce:
Closes #52

### DIFF
--- a/9.4-jre8/Dockerfile
+++ b/9.4-jre8/Dockerfile
@@ -1,0 +1,51 @@
+FROM openjdk:8-jre
+
+# add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
+RUN groupadd -r jetty && useradd -r -g jetty jetty
+
+ENV JETTY_HOME /usr/local/jetty
+ENV PATH $JETTY_HOME/bin:$PATH
+RUN mkdir -p "$JETTY_HOME"
+WORKDIR $JETTY_HOME
+
+ENV JETTY_VERSION 9.4.0.v20161208
+ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-distribution/$JETTY_VERSION/jetty-distribution-$JETTY_VERSION.tar.gz
+
+# GPG Keys are personal keys of Jetty committers (see https://dev.eclipse.org/mhonarc/lists/jetty-users/msg05220.html)
+ENV JETTY_GPG_KEYS \
+	# 1024D/8FB67BAC 2006-12-10 Joakim Erdfelt <joakime@apache.org>
+	B59B67FD7904984367F931800818D9D68FB67BAC \
+	# 1024D/D7C58886 2010-03-09 Jesse McConnell (signing key) <jesse.mcconnell@gmail.com>
+	5DE533CB43DAF8BC3E372283E7AE839CD7C58886
+RUN set -xe \
+	&& curl -SL "$JETTY_TGZ_URL" -o jetty.tar.gz \
+	&& curl -SL "$JETTY_TGZ_URL.asc" -o jetty.tar.gz.asc \
+	&& export GNUPGHOME="$(mktemp -d)" \
+	&& for key in $JETTY_GPG_KEYS; do \
+		gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$key"; done \
+	&& gpg --batch --verify jetty.tar.gz.asc jetty.tar.gz \
+	&& rm -r "$GNUPGHOME" \
+	&& tar -xvf jetty.tar.gz --strip-components=1 \
+	&& sed -i '/jetty-logging/d' etc/jetty.conf \
+	&& rm -fr demo-base javadoc \
+	&& rm jetty.tar.gz*
+
+ENV JETTY_BASE /var/lib/jetty
+RUN mkdir -p "$JETTY_BASE"
+WORKDIR $JETTY_BASE
+
+# Get the list of modules in the default start.ini and build new base with those modules, then add setuid
+RUN modules="$(grep -- ^--module= "$JETTY_HOME/start.ini" | cut -d= -f2 | paste -d, -s)" \
+	&& set -xe \
+	&& java -jar "$JETTY_HOME/start.jar" --add-to-startd="$modules,setuid"
+
+ENV TMPDIR /tmp/jetty
+RUN set -xe \
+	&& mkdir -p "$TMPDIR" \
+	&& chown -R jetty:jetty "$TMPDIR" "$JETTY_BASE"
+
+COPY docker-entrypoint.sh /
+
+EXPOSE 8080
+ENTRYPOINT ["/docker-entrypoint.sh"]
+CMD ["java","-jar","/usr/local/jetty/start.jar"]
--- a/9.4-jre8/Dockerfile
+++ b/9.4-jre8/Dockerfile
@@ -8,7 +8,7 @@ ENV PATH $JETTY_HOME/bin:$PATH
 RUN mkdir -p "$JETTY_HOME"
 WORKDIR $JETTY_HOME
 
-ENV JETTY_VERSION 9.4.0.v20161208
+ENV JETTY_VERSION 9.4.1.v20170120
 ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-home/$JETTY_VERSION/jetty-home-$JETTY_VERSION.tar.gz
 
 # GPG Keys are personal keys of Jetty committers (see https://dev.eclipse.org/mhonarc/lists/jetty-users/msg05220.html)
--- a/9.4-jre8/Dockerfile
+++ b/9.4-jre8/Dockerfile
@@ -33,10 +33,8 @@ ENV JETTY_BASE /var/lib/jetty
 RUN mkdir -p "$JETTY_BASE"
 WORKDIR $JETTY_BASE
 
-# Get the list of modules in the default start.ini and build new base with those modules, then add setuid
-RUN modules="$(grep -- ^--module= "$JETTY_HOME/start.ini" | cut -d= -f2 | paste -d, -s)" \
-	&& set -xe \
-	&& java -jar "$JETTY_HOME/start.jar" --add-to-startd="$modules,setuid"
+RUN set -xe \
+	&& java -jar "$JETTY_HOME/start.jar" --create-startd --add-to-start="server,http,deploy,jsp,jstl,ext,resources,websocket,setuid"
 
 ENV TMPDIR /tmp/jetty
 RUN set -xe \
--- a/9.4-jre8/Dockerfile
+++ b/9.4-jre8/Dockerfile
@@ -15,8 +15,8 @@ ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-distrib
 ENV JETTY_GPG_KEYS \
 	# 1024D/8FB67BAC 2006-12-10 Joakim Erdfelt <joakime@apache.org>
 	B59B67FD7904984367F931800818D9D68FB67BAC \
-	# 1024D/D7C58886 2010-03-09 Jesse McConnell (signing key) <jesse.mcconnell@gmail.com>
-	5DE533CB43DAF8BC3E372283E7AE839CD7C58886
+	# 4096R/438A3B7D 2016-08-15 Jesse McConnell (signing key) <jesse.mcconnell@gmail.com>
+	2A684B57436A81FA8706B53C61C3351A438A3B7D
 RUN set -xe \
 	&& curl -SL "$JETTY_TGZ_URL" -o jetty.tar.gz \
 	&& curl -SL "$JETTY_TGZ_URL.asc" -o jetty.tar.gz.asc \
--- a/9.4-jre8/Dockerfile
+++ b/9.4-jre8/Dockerfile
@@ -9,7 +9,7 @@ RUN mkdir -p "$JETTY_HOME"
 WORKDIR $JETTY_HOME
 
 ENV JETTY_VERSION 9.4.0.v20161208
-ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-distribution/$JETTY_VERSION/jetty-distribution-$JETTY_VERSION.tar.gz
+ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-home/$JETTY_VERSION/jetty-home-$JETTY_VERSION.tar.gz
 
 # GPG Keys are personal keys of Jetty committers (see https://dev.eclipse.org/mhonarc/lists/jetty-users/msg05220.html)
 ENV JETTY_GPG_KEYS \
@@ -27,7 +27,6 @@ RUN set -xe \
 	&& rm -r "$GNUPGHOME" \
 	&& tar -xvf jetty.tar.gz --strip-components=1 \
 	&& sed -i '/jetty-logging/d' etc/jetty.conf \
-	&& rm -fr demo-base javadoc \
 	&& rm jetty.tar.gz*
 
 ENV JETTY_BASE /var/lib/jetty
--- a/9.4-jre8/alpine/Dockerfile
+++ b/9.4-jre8/alpine/Dockerfile
@@ -18,8 +18,8 @@ ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-distrib
 ENV JETTY_GPG_KEYS \
 	# 1024D/8FB67BAC 2006-12-10 Joakim Erdfelt <joakime@apache.org>
 	B59B67FD7904984367F931800818D9D68FB67BAC \
-	# 1024D/D7C58886 2010-03-09 Jesse McConnell (signing key) <jesse.mcconnell@gmail.com>
-	5DE533CB43DAF8BC3E372283E7AE839CD7C58886
+	# 4096R/438A3B7D 2016-08-15 Jesse McConnell (signing key) <jesse.mcconnell@gmail.com>
+	2A684B57436A81FA8706B53C61C3351A438A3B7D
 
 RUN set -xe \
 	# Install required packages for build time. Will be removed when build finishes.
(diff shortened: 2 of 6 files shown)


hypnoce:
Tests are failing due to a ClassNotFoundException from the WebSocketClient class :
```
Caused by: 
java.lang.ClassNotFoundException: org.eclipse.jetty.client.HttpClient
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at org.eclipse.jetty.websocket.client.WebSocketClient.<init>(WebSocketClient.java:263)
```

I believe this is due to the recent introduction of a [dependency from websocket-client to jetty-client](https://github.com/eclipse/jetty.project/commit/e329e386f08552f55025f57f6c432c1562559dec#diff-aa3da4a6e8383e4808209667b1a92d3d) 
The easiest way to fix this problem is to make websocket.mod depend en client.mod.
It is also possible to add the necessary jars into lib/websocket folder.
Last solution is to declare the necessary jars into websocket.mod [lib] tag.
What do you think ?

md5:
@hypnoce Do you mind taking a look at the discussion in #34 and comparing your branch with the `jetty-9.4` branch I created a few months ago?

As for the `websocket.mod` issue you mention, perhaps @gregw or @jmcc0nn3ll might have some thoughts on that.

generate-stackbrew-library.sh
```diff
@@ -5,7 +5,7 @@ shopt -s globstar
 declare -A aliases
 aliases=(
 	[9.2-jre7]='jre7'
-	[9.3-jre8]='latest jre8'
+	[9.4-jre8]='latest jre8'
```
md5, line 8: I think if we're not going to have the Official Images team build a `jetty:9.3` image anymore, then it should be removed entirely from the repo.

@gregw Following up on our discussion in #34, which version of Jetty do you think we should provide up-to-date images of? Will you guys still be doing releases on the 9.3 line?

gregw: @md5, we have always run point  releases in parallel for a while and expect to do so for at least a little while.   So I would expect perhaps 1 or 2 more 9.3.x releases.   However as there are no technical barriers to upgrade to 9.4 (same JVM same API), then it will only those that need to port to changed APIs (sessions mostly) that may linger on 9.3

Note that in 9.4, we have added some features that make jetty easier to configure in a Dockerfile - better support for property substitution and the ability to put the confuguration options in a file rather than on the command line.   I'll have a look next week to see if there is scope to use those in 9.4, or at least include reference to them in the documentation.

gregw: Oh I should also say that we have a 9.3.x release coming out in the next week or two with some important bug fixes, so at the very least there should be at least one more 9.3.x docker release.

hypnoce:
@md5 Indeed, I see some difference regarding JETTY_RUN JETTY_STATE and adding the tmp dir to java global variable. I merged them.
Still waiting for the websocket.mod issue.
I had a look at the new jetty-home artifact. Here are the changes to use that new stripped down minimal version : 
https://github.com/hypnoce/docker-jetty/commit/bf3faf8603fc67a1299dd8d0b80e57601bb126fd
I can add them to the PR if necessary.

md5:
@hypnoce It actually looks like the `JETTY_RUN` stuff was removed in c228052ac9459c1303601e90fd66a44d7be6b8ce and never made it into the `jetty-9.4` branch. I think it's actually best for that part to align with `master`.

Glad to see the `jetty-home` artifact being used 👍 

Also, I realize now that I misinterpreted the changes to `generate-stackbrew-library.sh` since that change will just change the `latest` and `jre8` aliases to point to the 9.4 image, but the 9.3 image will still be included.

hypnoce:
@md5 I made the requested changes and used jetty-home.

Indeed _latest_ and _jre8_ are just aliases. So we can keep the 3 versions 9.2, 9.3, 9.4 and then remove/update them later on.

generate-stackbrew-library.sh
md5: Cool. What about the `websocket.mod` dependency issue?

md5:
@gregw What does the Jetty team recommend for a course of action for the `websocket.mod` dependency issue? I think it would be nice to get a Jetty 9.4 Docker image published.

generate-stackbrew-library.sh
jmcc0nn3ll: @joakime have you seen this websocket.mod issue?  I know you were working on leveraging jetty-client and seems like we need to add a dependency in the mod file.

joakime: Ugh!  Yes, this is a bug.

Opened eclipse/jetty.project#1183

joakime:
@md5 It was just fixed in eclipse/jetty.project#1183 , but in the meantime you can include `--add-to-start=client` as a temporary workaround until a new release is cut.

hypnoce:
So you prefer to force add to the start.ini for now and not to wait for the next release ?
@md5 either is fine.

md5:
@joakime What's your timeline for the next release?

jmcc0nn3ll:
looking like next week

hypnoce:
Hi,

Best wishes for this new year !

Any updates on the next release ?

Thanks

md5:
@jmcc0nn3ll @joakime Do you think 9.4.1 will be coming out this week? If not, it may be worthwhile for us to patch `websocket.mod` in the `Dockerfile` in order to get out a `jetty:9.4` image based on 9.4.0.

jmcc0nn3ll:
currently planning on staging it early next week

hypnoce:
Hi,

sorry for the delay. I inspected the error and apparently there is no more start.ini file in JETTY_HOME folder. Therefore, no module is loaded (no http module....) so jetty-hello-web test fails.
[RUN modules="$(grep -- ^--module= "$JETTY_HOME/start.ini" | cut -d= -f2 | paste -d, -s)"](https://github.com/appropriate/docker-jetty/pull/51/files#diff-dad6c4a2a53860754faaf15e201d8b0dR37)

Do you want to put all the previously loaded modules in the docker file ?
`RUN set -xe \
	&& java -jar "$JETTY_HOME/start.jar" --add-to-startd="home-base-warning,ext,server,jsp,resources,deploy,jstl,websocket,http,setuid"`

Thanks

joakime:
The modification of `${jetty.home}` is discouraged.

https://www.eclipse.org/jetty/documentation/current/startup-base-and-home.html

You are meant to setup your own `${jetty.base}` directory, with the modules that you need/want.
Also, do not use `home-base-warning` in your jetty.base directory (that merely warns you to not modify the jetty.home directory)

hypnoce:
We are not modifying jetty_home here but adding modules to jetty_base. Here I was just proposing to enable the same modules as in Jetty 9.3 in the new jetty 9.4 image so that users do not get lost when updating.

md5:
The Dockerfile isn't modifying JETTY_HOME, it's provisioning a new JETTY_BASE. It previously used the list in start.ini as the starting point for “default” modules, but it seems that's no longer possible with the new minimal distribution artifact.

Is there some other place to get a reasonable list of default modules for this image? I don't think it makes sense to hard code it in the Dockerfile.

joakime:
Ah, understood.

The problem now is that there's no such thing as "default" for a Jetty distribution.
Its all dependent on the `jetty.base` directory and its configuration.

Even before you did this change to use the `jetty-home.tar.gz` that was true.
The old `jetty-distribution.tar.gz` had a `start.ini`, that's true, but that file is not used for a `jetty.base` directory or the eventual instance configuration.

Whatever configuration you were using before for your `jetty.base` directory (before the change from `jetty-distribution.tar.gz` to `jetty-home.tar.gz`) should still be valid.

md5:
This image has always used the `start.ini` in `JETTY_HOME` as a starting point for adding modules to a default `JETTY_BASE`. I suppose that may have been a mistake on my part, but it seemed to provide a useful default for users who just want to add their WAR file and run an application. I'm not sure that the alternative of having an empty `JETTY_BASE` or having no `JETTY_BASE` at all would have been resulted in a usable image.

At this point, it seems like we can either go back to using the `jetty-distribution` artifact as a starting point or start hard-coding our own list of modules to be added to `JETTY_BASE` by default.

@joakime: is the latter what you mean by this:

> Whatever configuration you were using before for your jetty.base directory (before the change from jetty-distribution.tar.gz to jetty-home.tar.gz) should still be valid.

This is basically what @hypnoce was proposing, aside from our not needing to include `home-base-warning`. If that's the best we can do to ensure consistency for users moving from 9.3 to 9.4, I guess that's better than breaking the image entirely for `jetty:9.4`.

gregw:
I think that it is reasonable to just code a minimal set of modules that will allow a webapp to be deployed and a connector to list for request.

Absolute minimum is `server`, `http`,`deploy`

To be compatible with previous images we can add: `jsp`, `jstl` and `websocket` as enabled features and `ext` and `resources` as easy ways for derived images to extend the classpath.

9.4-jre8/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+	&& rm jetty.tar.gz*
+
+ENV JETTY_BASE /var/lib/jetty
+RUN mkdir -p "$JETTY_BASE"
+WORKDIR $JETTY_BASE
+
+# Get the list of modules in the default start.ini and build new base with those modules, then add setuid
+RUN modules="$(grep -- ^--module= "$JETTY_HOME/start.ini" | cut -d= -f2 | paste -d, -s)" \
+	&& set -xe \
+	&& java -jar "$JETTY_HOME/start.jar" --add-to-startd="$modules,setuid"
```
gregw: `--add-to-startd=modlist`  is deprecated.
Instead you should have a `--create-startd --add-to-start=modlist`

9.4-jre8/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+	&& tar -xvf jetty.tar.gz --strip-components=1 \
+	&& sed -i '/jetty-logging/d' etc/jetty.conf \
+	&& rm jetty.tar.gz*
+
+ENV JETTY_BASE /var/lib/jetty
+RUN mkdir -p "$JETTY_BASE"
+WORKDIR $JETTY_BASE
+
+# Get the list of modules in the default start.ini and build new base with those modules, then add setuid
+RUN modules="$(grep -- ^--module= "$JETTY_HOME/start.ini" | cut -d= -f2 | paste -d, -s)" \
```
gregw: modules="server,http,deploy,jsp,jstl,ext,resources"  is a reasonable default.

9.4-jre8/alpine/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+	&& rm -r "$GNUPGHOME" \
+	&& tar -xvzf jetty.tar.gz \
+	&& mv jetty-home-$JETTY_VERSION/* ./ \
+	&& sed -i '/jetty-logging/d' etc/jetty.conf \
+	&& rm jetty.tar.gz* \
+	&& rm -fr jetty-home-$JETTY_VERSION/ \
+
+	# Get the list of modules in the default start.ini and build new base with those modules, then add setuid
+	&& cd $JETTY_BASE \
+	&& modules="$(grep -- ^--module= "$JETTY_HOME/start.ini" | cut -d= -f2 | paste -d, -s)" \
```
md5: Looks like we should drop this `modules` line as well as the comment above, right?

9.4-jre8/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+	&& rm -r "$GNUPGHOME" \
+	&& tar -xvf jetty.tar.gz --strip-components=1 \
+	&& sed -i '/jetty-logging/d' etc/jetty.conf \
+	&& rm jetty.tar.gz*
+
+ENV JETTY_BASE /var/lib/jetty
+RUN mkdir -p "$JETTY_BASE"
+WORKDIR $JETTY_BASE
+
+# Get the list of modules in the default start.ini and build new base with those modules, then add setuid
```
md5: Let's drop this comment, since it's no longer accurate.

md5 (changes requested):

md5:
Thanks for the feedback @joakime and @gregw.

@hypnoce I think we can merge this once my latest comments are addressed.

9.4-jre8/alpine/Dockerfile
hypnoce: Sorry my bad

hypnoce:
Everything should be ok now. Thanks for your comments.

md5:
Diff between 9.3 and 9.4 looks good to me:

```diff
diff -ru 9.3-jre8/Dockerfile 9.4-jre8/Dockerfile
--- 9.3-jre8/Dockerfile	2017-02-04 07:40:10.000000000 -0800
+++ 9.4-jre8/Dockerfile	2017-02-04 07:40:43.000000000 -0800
@@ -8,15 +8,15 @@
 RUN mkdir -p "$JETTY_HOME"
 WORKDIR $JETTY_HOME
 
-ENV JETTY_VERSION 9.3.12.v20160915
-ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-distribution/$JETTY_VERSION/jetty-distribution-$JETTY_VERSION.tar.gz
+ENV JETTY_VERSION 9.4.1.v20170120
+ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-home/$JETTY_VERSION/jetty-home-$JETTY_VERSION.tar.gz
 
 # GPG Keys are personal keys of Jetty committers (see https://dev.eclipse.org/mhonarc/lists/jetty-users/msg05220.html)
 ENV JETTY_GPG_KEYS \
 	# 1024D/8FB67BAC 2006-12-10 Joakim Erdfelt <joakime@apache.org>
 	B59B67FD7904984367F931800818D9D68FB67BAC \
-	# 1024D/D7C58886 2010-03-09 Jesse McConnell (signing key) <jesse.mcconnell@gmail.com>
-	5DE533CB43DAF8BC3E372283E7AE839CD7C58886
+	# 4096R/438A3B7D 2016-08-15 Jesse McConnell (signing key) <jesse.mcconnell@gmail.com>
+	2A684B57436A81FA8706B53C61C3351A438A3B7D
 RUN set -xe \
 	&& curl -SL "$JETTY_TGZ_URL" -o jetty.tar.gz \
 	&& curl -SL "$JETTY_TGZ_URL.asc" -o jetty.tar.gz.asc \
@@ -27,17 +27,14 @@
 	&& rm -r "$GNUPGHOME" \
 	&& tar -xvf jetty.tar.gz --strip-components=1 \
 	&& sed -i '/jetty-logging/d' etc/jetty.conf \
-	&& rm -fr demo-base javadoc \
 	&& rm jetty.tar.gz*
 
 ENV JETTY_BASE /var/lib/jetty
 RUN mkdir -p "$JETTY_BASE"
 WORKDIR $JETTY_BASE
 
-# Get the list of modules in the default start.ini and build new base with those modules, then add setuid
-RUN modules="$(grep -- ^--module= "$JETTY_HOME/start.ini" | cut -d= -f2 | paste -d, -s)" \
-	&& set -xe \
-	&& java -jar "$JETTY_HOME/start.jar" --add-to-startd="$modules,setuid"
+RUN set -xe \
+	&& java -jar "$JETTY_HOME/start.jar" --create-startd --add-to-start="server,http,deploy,jsp,jstl,ext,resources,websocket,setuid"
 
 ENV TMPDIR /tmp/jetty
 RUN set -xe \
diff -ru 9.3-jre8/alpine/Dockerfile 9.4-jre8/alpine/Dockerfile
--- 9.3-jre8/alpine/Dockerfile	2017-02-04 07:40:10.000000000 -0800
+++ 9.4-jre8/alpine/Dockerfile	2017-02-04 07:40:43.000000000 -0800
@@ -11,15 +11,15 @@
 ENV JETTY_BASE /var/lib/jetty
 RUN mkdir -p "$JETTY_BASE"
 
-ENV JETTY_VERSION 9.3.12.v20160915
-ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-distribution/$JETTY_VERSION/jetty-distribution-$JETTY_VERSION.tar.gz
+ENV JETTY_VERSION 9.4.1.v20170120
+ENV JETTY_TGZ_URL https://repo1.maven.org/maven2/org/eclipse/jetty/jetty-home/$JETTY_VERSION/jetty-home-$JETTY_VERSION.tar.gz
 
 # GPG Keys are personal keys of Jetty committers (see https://dev.eclipse.org/mhonarc/lists/jetty-users/msg05220.html)
 ENV JETTY_GPG_KEYS \
 	# 1024D/8FB67BAC 2006-12-10 Joakim Erdfelt <joakime@apache.org>
 	B59B67FD7904984367F931800818D9D68FB67BAC \
-	# 1024D/D7C58886 2010-03-09 Jesse McConnell (signing key) <jesse.mcconnell@gmail.com>
-	5DE533CB43DAF8BC3E372283E7AE839CD7C58886
+	# 4096R/438A3B7D 2016-08-15 Jesse McConnell (signing key) <jesse.mcconnell@gmail.com>
+	2A684B57436A81FA8706B53C61C3351A438A3B7D
 
 RUN set -xe \
 	# Install required packages for build time. Will be removed when build finishes.
@@ -33,16 +33,12 @@
 	&& gpg --batch --verify jetty.tar.gz.asc jetty.tar.gz \
 	&& rm -r "$GNUPGHOME" \
 	&& tar -xvzf jetty.tar.gz \
-	&& mv jetty-distribution-$JETTY_VERSION/* ./ \
+	&& mv jetty-home-$JETTY_VERSION/* ./ \
 	&& sed -i '/jetty-logging/d' etc/jetty.conf \
-	&& rm -fr demo-base javadoc \
 	&& rm jetty.tar.gz* \
-	&& rm -fr jetty-distribution-$JETTY_VERSION/ \
-
-	# Get the list of modules in the default start.ini and build new base with those modules, then add setuid
+	&& rm -fr jetty-home-$JETTY_VERSION/ \
 	&& cd $JETTY_BASE \
-	&& modules="$(grep -- ^--module= "$JETTY_HOME/start.ini" | cut -d= -f2 | paste -d, -s)" \
-	&& java -jar "$JETTY_HOME/start.jar" --add-to-startd="$modules,setuid" \
+	&& java -jar "$JETTY_HOME/start.jar" --create-startd --add-to-start="server,http,deploy,jsp,jstl,ext,resources,websocket,setuid" \
 
 	# Remove installed packages and various cleanup
 	&& apk del .build-deps \
```

md5 (approved):

md5:
I opened a request to add the new image here: https://github.com/docker-library/official-images/pull/2618

Thanks for your work on this and your patience @hypnoce! :dancing_men: 👯  🎉

PR #62, opened 2017-05-10
Fix #61 Provide Jetty-images with JDK

### LINKED ISSUE #61: Provide Jetty-images with JDK instead of JRE
Hi,

thanks for providing docker-Jetty-images.
In my use case, I do not only want to use Jetty, I have to use JDK (instead of JRE), too¹. 
It would be great, if you could provide additional docker-Jetty-images, which contain JDK instead of JRE.

¹ 
In special: Within my docker-image, I want to be able to use [v8](https://developers.google.com/v8/). And I want to have [Alpine](https://hub.docker.com/_/alpine/) as base image. Because of Alpine, I have to compile v8 on my own, needing JDK at this point, for details see https://github.com/tarent/openjdk-alpine-j2v8. So I want Jetty, Alpine, and am needing JDK for compiling v8. 
I want to adapt the approach of tarent/openjdk-alpine-j2v8 (which extends openjdk:8-jdk-alpine) to the Jetty docker image. For this, I am "missing" jetty:jdk8-alpine.

SchulteDev:


gregw:
This is a moderately high maintenance path to provide the alternate images for JDK. 

So firstly, can you explain your use-case a bit more for why you need the JDK rather than JRE?   Assuming a reasonable answer to that, I think a better approach may be to adopt an approach where the one maven module is parameterized so the Dockerfile is templated out with either `FROM openjdk-jreX` or `FROM openjdk-jdkX`

SchulteDev:
This PR was only a Quickwin-approach. Let's have the discussion at #61 .

PR #64, opened 2017-05-24
Adding my GPG keys

gregw:
It's been a while since I ran the jetty release myself, but not an impossibility going forward


md5:
This looks good to merge if you rebase out the merge commit and get Travis to pass. Looks like you need to publish your keys maybe?

Issue #63, opened 2017-05-11
Bypassing start.jar when starting Jetty


SchulteDev:
Hi,

Jetty's start-time can be decreased by [bypassing start.jar](https://www.eclipse.org/jetty/documentation/9.4.x/quickstart-webapp.html#_bypassing_start_jar).

Please add this improvement to the docker-jetty images.

gregw:
The problem with baking in the `start.jar` bypass into a base docker image, is that extended images will need to regenerate the saved commands whenever they change the jetty configuration.

It may be preferable to simply document how it can be done in a extended docker image rather than put something in the base.

I'll do that at least for a start...

SchulteDev:
Maybe a mix? Let's say, a predefined command, which I can use in my extended docker image, or a script within the docker image for re-usage; combined with some documentation.

md5:
@SchulteMarkus I agree with @gregw that baking this into the image is problematic. I also don't see it being much help to have a predefined command when you can pretty easily do something like this:

```Dockerfile
FROM jetty:9.4

RUN \
	echo "#!/bin/sh" > /jetty-quick-start \
	&& echo "exec $(java -jar $JETTY_HOME/start.jar --dry-run)" >> /jetty-quick-start \
	&& chmod +x /jetty-quick-start

ENTRYPOINT ["/jetty-quick-start"]
```

SchulteDev:
I would prefer not to overwrite ENTRYPOINT from jetty-image

gregw:
@SchulteMarkus the entry point is already a rather complex script and I'm fearful that it could be very confusing if we added quickstart support.

For example, let's say we make it auto-detect a quick-start.sh script and run it if it exists, then what should the entry point do if it passed extra argument which it would normally add to the command line passed to start.jar.

I can see somebody trying to pass args to such and image and getting very confused as to why they are not being used.

I think replacing the entrypoint is a good clear way to indicate an image is starting differently.  It's also quicker than calling a script that calls a script.

Conversely with the example posted by @md5 I can see users getting confused trying to pass in 'bash' as an argument, but having it ignored and jetty started anyway.

So perhaps the compromise is to provide a command in an image that can be run by a Dockerfile that generates a quick-start-entrypoint that does the right thing with extra args and commands, but avoids start.jar.  It could even perhaps check date stamps on the generated script vs the jetty start.d directory to warn if jetty's configuration has been changed since the script was generated.

SchulteDev:
[...]

PR #53, opened 2016-12-22
Update to Jetty 9.2.20.v20161216 and 9.3.15.v20161220

md5:

PR #66, opened 2017-05-24
Issue #63 Bypass start.jar

gregw:
While this is working, I'm not sure this PR is ready to merge yet, I just wanted to create it as the basis of discussion.

This modified docker-entrypoint.sh , so that after the args have been processed, it looks to see if it is a jetty command and if so:
 * If it has a terminating argument (eg --dry-run or --version), then just exec the command
 * Else If a `/jetty-quickstart` files exists and is newer than `$JETTY_BASE/start.d` then set the contents of that file as the command
 * Else set the command as the output of the jetty command with --dry-run appended

This approach:
 * always avoid spawning a nested JVM and the main jetty process is PID 1.   
 * allows a pre-generated jetty-quickstart file to bypass running start.jar at all
 * still supports passing in args and/or commands and acts as you would expect

Some things I'm not really happy with:
 * my sh-foo is not so great, so this could perhaps be done more simply... and I'm a bit worried about any args that have spaces in them.
 * listing the terminating arguments is a little fragile
 * on reflection the file should be called something like `/jetty-command` as it is unrelated to quickstart
 * perhaps we should warn if a `/jetty-command` exists but is older than `start.d`?

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -65,11 +65,22 @@ if expr "$*" : '^java .*/start\.jar.*$' >/dev/null ; then
 		esac
 	done
 
-	if [ -f /jetty-quickstart -a /jetty-quickstart -nt $JETTY_BASE/start.d ] ; then
-		set -- $(cat /jetty-quickstart)
+	if [ -f /jetty-start ] ; then
+		if [ $JETTY_BASE/start.d -nt /jetty-start ] ; then
+			cat >&2 <<- 'EOWARN'
+			********************************************************************
+			WARNING: The $JETTY_BASE/start.d directory has been modified since
+			         the /jetty-start files was generated. Please either delete 
+			         the /jetty-start file or re-run /generate-jetty-start.sh 
+			         from a Dockerfile
+			********************************************************************
+			EOWARN
+		fi
+		echo $(date +'%Y-%m-%d %H:%M:%S.000'):INFO:docker-entrypoint:jetty start command from /jetty-start
+		set -- $(cat /jetty-start)
 	else
 		# Do a jetty dry run to set the final command
-		set -- $("$@" --dry-run)
+		set -- $("$@" --dry-run | sed 's/\\$//' )
 	fi
 fi
 
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -38,4 +38,39 @@ if [ "$1" = "java" -a -n "$JAVA_OPTIONS" ] ; then
 	set -- java $JAVA_OPTIONS "$@"
 fi
 
+if expr "$*" : '^java .*/start\.jar.*$' >/dev/null ; then
+	# this is a command to run jetty
+
+	# check if it is a terminating command
+	for A in "$@" ; do
+		case $A in
+			--add-to-start* |\
+			--create-files |\
+			--create-startd |\
+			--download |\
+			--dry-run |\
+			--exec-print |\
+			--help |\
+			--info |\
+			--list-all-modules |\
+			--list-classpath |\
+			--list-config |\
+			--list-modules* |\
+			--stop |\
+			--update-ini |\
+			--version |\
+			-v )\
+			# It is a terminating command, so exec directly
+		        exec "$@"
+		esac
+	done
+
+	if [ -f /jetty-quickstart -a /jetty-quickstart -nt $JETTY_BASE/start.d ] ; then
+		set -- $(cat /jetty-quickstart)
+	else
+		# Do a jetty dry run to set the final command
+		set -- $("$@" --dry-run)
+	fi
+fi
+
 exec "$@"
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -38,7 +38,7 @@ if [ "$1" = "java" -a -n "$JAVA_OPTIONS" ] ; then
 	set -- java $JAVA_OPTIONS "$@"
 fi
 
-if expr "$*" : '^java .*/start\.jar.*$' >/dev/null ; then
+if expr "$*" : 'java .*/start\.jar.*$' >/dev/null ; then
 	# this is a command to run jetty
 
 	# check if it is a terminating command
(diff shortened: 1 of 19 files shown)


gregw:
For Issue #63

gregw:
@md5 @SchulteMarkus your thoughts?

It's a little bit more complexity than I'd like in the entrypoint, but I think it works well enough and has some good benefits.

SchulteDev:
I think this is the way to go for solving the issue. There is more complexity, and I agree, it is kind of fragile (what if Jetty adds a new program-argument?), but I think, this is worth it, as many users will benefit.

md5:
I'm not sure that the complexity here is worth the benefit, but I'm to concede that the benefit may be greater than I perceive.

From my testing, the Jetty process is already PID 1 with the default behavior of this image. There may be some overhead associated with setting up a new `ClassLoader` and executing the `main` method of the Jetty server, but that complexity is hidden from the user. Only when the `--exec` argument is used do I see a case where the Jetty server is a different PID than 1.

The second listed benefit is that a pre-generated script can be used to avoid `start.jar` entirely, but this benefit can already be realized without adding any complication to the entrypoint. The `generate-jetty-start.sh` script would look substantially the same either way.

The only benefit that remains is the ability to pass command line arguments to the container invocation while still retaining the ability to avoid `start.jar`, but I find this benefit to be dubious in the face of the complexity introduced. In my mind, if you're trying to shave milliseconds (or perhaps a second) off the startup time, it's worth having your own entrypoint as well.

That being said, I don't want to stand in the way of the Jetty team taking ownership over the direction of this image. I think the Shell scripting looks fine and appears to work as expected on Alpine as well as Debian. I agree that the terminating arguments list is fragile, but it seems like the sort of thing that would break infrequently and in hopefully obvious ways.

gregw:
@md5 Mike,

I too am a little cautious about the approach.   However, what I do like is that if we are to encourage more users to use quickstart, then we will retain a single entrypoint and a single approach.  So having initially been sceptical that this was the right thing, I find myself now probably on the side that it is a good thing.

note that it is not only `--exec` that will cause the behaviour, but enabling any module that has system properties or jvm args also has an implicit `--exec`.   In my work with other docker images there has definitely been confused/surprised users when enabling a module (eg Java Util Logging) caused the PID to no longer be 1.

With regards to processing command line args - I don't think that is a huge benefit other than to maintain the standard behaviour that args can be passed in.  It is again about not surprising users rather than any huge technical benefit.

So I currently think this is the right way to go, However, I wont merge just yet.   I'll ponder and test it a bit more just to be really sure there is not a better way.  I'll comment again before taking any action either way,

cheers

md5:
Sounds good 👍

md5 (approved):

gregw:
@md5 I've worked on this PR a little bit to make it more robust if we are not rigorous in maintaining the list of terminating options.  If fact we could almost get by without the list or at least a very minimal version of exceptional cases.

I'd like to merge this soon, so can you double check that you are still OK with it.

docker-entrypoint.sh
```diff
@@ -80,7 +80,16 @@ if expr "$*" : '^java .*/start\.jar.*$' >/dev/null ; then
 		set -- $(cat /jetty-start)
 	else
 		# Do a jetty dry run to set the final command
-		set -- $("$@" --dry-run | sed 's/\\$//' )
+		"$@" --dry-run > /jetty-start
+		EXIT=$?
```
md5: Since `set -e` is in effect here, won't this always be zero?

docker-entrypoint.sh
```diff
@@ -80,7 +80,16 @@ if expr "$*" : '^java .*/start\.jar.*$' >/dev/null ; then
...
+		if [ $(egrep -v '\\$' /jetty-start | wc -l ) -gt 1 ] ; then
```
md5: What's the intent of this `'\\$'`? It seems like the `awk` invocation below is assuming it's the first character, so should it be anchored with `^`?

gregw: The annoying thing is that jetty-9.2 --dry-run doesn't produce a single line. Instead it produces multiple lines split with a final `\` character.  So this egrep to wc is counting the number of lines that do not terminate with `\` (and specifically it is not anchered with `^`).

docker-entrypoint.sh
gregw: Good point.   But if it exits, we lose any stdout that might explain why... although I guess it should put any errors onto stderr.

Two choices:
a) don't capture the exit code and just exit 0 if there are more than 1 lines in the output.
b) turn off -e

I'm good with a) because looking at start.jar code, it appears to write to stderr before any non zero exit

md5: I would probably do this:

```shell
"$@" --dry-run > /jetty-start || EXIT=$?
```

That way a non-zero exit code won't cause that line to abort and will capture the exit code. In the case of a zero exit code, `$EXIT` will remain uninitialized, which I think is fine.

md5: It's good practice to keep `set -e` globally in case we introduce any bugs later where we're not properly accounting for failed commands.

docker-entrypoint.sh
md5: Makes sense. I totally misread that regexp. I thought the `\\` was escaping a backslash at the shell level and that the effective regexp was `/\$/`, not `/\\$/`.

md5:
Looks like some bad rebasing snuck in here? I think it would be good to squash and rebase these changes into a single commit on top of 9c03b274e981c7405b956f6f158e8694c55d0838 before merging, once we settle on the `set -e` thing.

gregw:
@md5 The reference to #53 was just me having finger troubles typing commit comments.  So I repeated all the comments in my final commit.

I'm happy with retaining the -e and only doing a log and exit 0 for additional stdout lines.  The branch is like that now, so I'm good to squash/merge  and correct the comments.

I'll give it another 24h before doing so, so you can comment if there is something I missed.

md5:
Looks good to me 👍

Phocea:
Hello,

Do you know when this merge is going to also happen on the [Alpine version](https://github.com/appropriate/docker-jetty/blob/bb1a968360166dec223b0eff1a979e1acad8b976/9.4-jre8/alpine/Dockerfile) please ?

Thanks

gregw:
@Phocea this PR did leave out the generate script from the alpine Dockerfile, but it has [subsequently been fixed](https://github.com/appropriate/docker-jetty/blob/master/9.4-jre8/alpine/Dockerfile#L69) and the master branch is up to date.  

I will look at updating the official images as soon as I get an opportunity.

Phocea:
Great. Thanks for the update

gregw:
official images update has been merged.

PR #67, opened 2017-06-07
Update Jetty versions

md5:
Update to Jetty 9.2.22, 9.3.20, and 9.4.6


md5:
We were seeing a weird failure during the `rm -r "$GNUPGHOME"` step that was happening only on Alpine, though I wasn't able to reproduce it locally. I changed it to `rm -rf "$GNUPGHOME"`, which resolved the issue.

PR #71, opened 2017-08-17
Added quick test to check all images

gregw:
This is a simple script that runs up each image, checking that it accepts arguments, reports the actual jetty version and hits a simple page over HTTP.

### DIFF
--- a/quick-test.sh
+++ b/quick-test.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+
+PORT=8888
+
+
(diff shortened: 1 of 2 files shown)


md5:
Have a look at the test suite we use both for this repo (via Travis) and for the Official Images build process: https://github.com/docker-library/official-images/tree/master/test/tests/jetty-hello-web

Maybe it makes more sense to enhance those tests than to add additional tests in this repo.

gregw:
Perhaps, although it is very convenient to have a quick test within the project.  However, I can easily move the script to my own local bin and it need not be part of the repo if there are tests elsewhere.

Issue #70, opened 2017-08-17
Avoid setuid usage


gregw:
setuid usage requires native libraries that reduces the architectures that can be supported out of the box.  We should avoid setuid and document how it can be used if desired. (see discussion in #68)

PR #72, opened 2017-08-17
No setuid

gregw:
The technical side of #70 .  Will also produce a PR for the documentation soon.

Note this PR branched off the quicktest one.  If that's not acceptable, I'll remake this one.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -26,12 +26,14 @@ if ! command -v -- "$1" >/dev/null 2>&1 ; then
 	set -- java -jar "$JETTY_HOME/start.jar" "$@"
 fi
 
-if [ -n "$TMPDIR" ] ; then
-	case "$JAVA_OPTIONS" in
-		*-Djava.io.tmpdir=*) ;;
-		*) JAVA_OPTIONS="-Djava.io.tmpdir=$TMPDIR $JAVA_OPTIONS" ;;
-	esac
+if [ -z "$TMPDIR" ] ; then
+	TMPDIR=/tmp/jetty
+	mkdir $TMPDIR 2>/dev/null
 fi
+case "$JAVA_OPTIONS" in
+	*-Djava.io.tmpdir=*) ;;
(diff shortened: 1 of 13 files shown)


gregw:
I've removed the quicktest

docker-entrypoint.sh
```diff
@@ -80,15 +80,15 @@ if expr "$*" : 'java .*/start\.jar.*$' >/dev/null ; then
...
+		if [ $(egrep -v '\\$' /tmp/jetty-start | wc -l ) -gt 1 ] ; then
 			# command was more than a dry-run
-			cat /jetty-start \
+			cat /tmp/jetty-start \
 			| awk '/\\$/ { printf "%s", substr($0, 1, length($0)-1); next } 1' \
 			| egrep -v '[^ ]*java .* org\.eclipse\.jetty\.xml\.XmlConfiguration '
 			exit
 		fi
-		set -- $(sed 's/\\$//' /jetty-start)
+		set -- $(sed 's/\\$//' /tmp/jetty-start)
```
md5: What's the `/tmp` change about?

gregw: The entrypoint no longer runs as root, so it can't create the on the fly version of /jetty-start in root.  So just create it in /tmp instead.

md5: Maybe it would make sense to `touch` and `chown` the file at `/jetty-start` and add an extra check for a non-zero with `test -s`? I'm leery of using `/tmp` since it's a prime candidate for mounting into the container with `tmpfs`.

md5: Either that or put it under `$JETTY_BASE`, which is already owned by `jetty`. Though now that I look, I see we still have code setting `TMPDIR` to `/tmp/jetty` and `chown`'ing that...

gregw:
@md5 I think this is good to go, the `/tmp` usage has been moved to `$TMPDIR` which is by default `/tmp/jetty`

md5 (approved):

nafg:
Where's the documentation?

gregw:
Sorry, I'm behind on looking at the documentation.  Will get to it ASAP.

PR #73, opened 2017-09-20
update to jetty-9.4.7

gregw:


gregw:
Hmmm that failure looks to be a certificate lookup issue?

md5 (approved):
I'm re-running the failed job now

md5:
@gregw Do you want to open a PR against https://github.com/docker-library/official-images to get this released?

md5:
@gregw Am I imagining things or did you post a comment on this thread about waiting for 9.4.8? Maybe it was just via email, but I can't find your message there either. Maybe I'm losing it...

gregw:
@md5 It was via email.  We could promote 9.4.7, but it has several reports of performance degradation, although that is not necessary reproducible and only for machines with >32 CPUs.   9.4.8 is probably several weeks away, so we may as well release 9.4.7 as it does have a fair bit of goodness there as well.

gregw:
@md5 actually, we've just done a 9.3 release, so I'll update that and then get both in the official images PR

gregw:
now it is me getting confused... I've already done that!

md5:
@gregw Looks like the 9.3.21 upgrade didn't make it into the `official-images` PR, despite being merged here as part of this one.

Did you run `generate-stackbrew-library.sh` to create your PR against `official-images`?

gregw:
Brain fade sorry... I saw I had already done the 9.3 image update, but then didn't update the PR (which I generated manually).  Will give the script a go now...

gregw:
@md5   https://github.com/docker-library/official-images/pull/3508
This also caught the changes made to the entry point for setuid and avoiding JVM forks for 9.2

PR #78, opened 2017-09-28
Fix handling of JAVA_OPTIONS in entrypoint

### LINKED ISSUE #76: JAVA_OPTIONS not working
It would seem with the latest release of jetty:9-alpine and jetty:latest `JAVA_OPTIONS` set via environment variable is not working.

Simple test case:
```
docker run --rm --name jettymem -ti -e JAVA_OPTIONS=-Xmx1G jetty:latest
```

Check command run with `docker top jettymem`, the -Xmx1G is not present even though you can see it is set with `docker exec -ti jettymem sh -c ' echo "$JAVA_OPTIONS" '`

md5:
Fixes #76

### DIFF
--- a/9.2-jre7/docker-entrypoint.sh
+++ b/9.2-jre7/docker-entrypoint.sh
@@ -35,11 +35,6 @@ case "$JAVA_OPTIONS" in
 	*) JAVA_OPTIONS="-Djava.io.tmpdir=$TMPDIR $JAVA_OPTIONS" ;;
 esac
 
-if [ "$1" = "java" -a -n "$JAVA_OPTIONS" ] ; then
-	shift
-	set -- java $JAVA_OPTIONS "$@"
-fi
-
 if expr "$*" : 'java .*/start\.jar.*$' >/dev/null ; then
 	# this is a command to run jetty
 
@@ -92,6 +87,12 @@ if expr "$*" : 'java .*/start\.jar.*$' >/dev/null ; then
 		fi
 		set -- $(sed 's/\\$//' $TMPDIR/jetty-start)
 	fi
+fi
+
+if [ "${1##*/}" = java -a -n "$JAVA_OPTIONS" ] ; then
+	java="$1"
+	shift
+	set -- "$java" $JAVA_OPTIONS "$@"
 fi
 
 exec "$@"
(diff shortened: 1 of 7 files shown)


md5:
It's not clear to me what the Travis failure is about. Both failures are on the Alpine images and look similar, but they're failing on different tests. I think I'll restart them and see what happens.

9.2-jre7/docker-entrypoint.sh
```diff
@@ -94,4 +89,10 @@ if expr "$*" : 'java .*/start\.jar.*$' >/dev/null ; then
 	fi
 fi
 
+if [[ ($1 = java || $1 == */java) && -n "$JAVA_OPTIONS" ]] ; then
+	java="$1"
+	shift
+	set -- "$java" $JAVA_OPTIONS "$@"
```
gregw: The only problem I can think of moving this to here is that the JAVA_OPTIONs are not applied to the runs of java that occur at line 61 and 80.

gregw (changes requested):
I think that JAVA_OPTIONS needs to be applied at least at  line 61, probably also line 80

9.2-jre7/docker-entrypoint.sh
md5: Can you think of any JVM options that would be relevant to the terminating commands that get executed on that line?

I suppose it wouldn't hurt to just leave the old setting of `JAVA_OPTIONS` in place and to also add this one here.

md5:
I think I fixed the build issue with Alpine. The problem is that it doesn't have `/bin/bash`. I've used a shell substitution to strip any leading directory name from `$1` to check if the command is `java`.

9.2-jre7/docker-entrypoint.sh
gregw: i'm struggling to think of a really valid use-case. perhaps debugging when you might want to pass -verbose.  So not a show stopper.

md5: Let's just go with dealing with `$JAVA_OPTIONS` in one place. We can always add it back later if we need to, but I don't think this will break anything (famous last words).

gregw: ack

Issue #80, opened 2017-09-29
USER jetty + /generate-jetty-start.sh = Permission denied


jeremy-im:
`/generate-jetty-start.sh: 3: /generate-jetty-start.sh: cannot create /jetty-start: Permission denied`

md5:
Can you post your `Dockerfile`?

It looks like you need to temporarily switch to `USER root` for that command and then switch back to `USER jetty`.

gregw:
@md5 This looks like it is going to be a common issue.  Perhaps we should put some explicit code into the script so that it gives a reasonable warning and explanation that USER needs to be switched back to ROOT.   We should then probably warn in the entry point if the USER is not switched back to `jetty`

md5:
@gregw Sure, that sounds good. I think it would also be good to put an example in the documentation.

gregw:
@md5 actually switching USER frequently is against [best practises](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#user).  Perhaps we'd be better off moving `/jetty-start' into `$JETTY_BASE` so it can be written without changing users?

I'll prepare a PR for that....

md5:
I'm not sure the advice against switching `USER` frequently really applies here. We only switch it once in the `Dockerfile` for this image and I think having derived images switch to `USER root` to run commands that need `root` before going back to `USER jetty` isn't too bad. I suppose it could get a little out of hand if you had multiple layers of derived images.

As for whether it's actually a good practice to have `jetty.start` owned by `jetty`, I would argue that having all files that should not be writable by the `jetty` user from the running application owned by `root` would be more secure. This would help prevent attackers from gaining a foothold inside a long-running container by modifying files inside it. Using Docker's read-only mode goes one step further.

jeremy-im:
Aplogies for my terseness in opening this issue. But you got to an understanding of my complaint in spite of that.

For my two cents, I think this is mostly about documentation. An example would cover it. And the README shown at hub.docker.com needs updating. Thanks!

gregw:
@md5, I'm not sure this is a security issue as anybody that can write to `$JETTY_BASE` has complete control of the server anyway as they can replace any jars etc.  

[...]

PR #81, opened 2017-10-01
Issue #80 generate-jetty-start.sh Permission denied

gregw:
Fixes for #80 

Moved the /jetty-start file to $JETTY_BASE/jetty.start, where it can be
written to by the jetty user.

Added a warning in the docker-entrypoint.sh if the user is not 'jetty'

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -62,30 +62,40 @@ if expr "$*" : 'java .*/start\.jar.*$' >/dev/null ; then
 		esac
 	done
 
-	if [ -f /jetty-start ] ; then
-		if [ $JETTY_BASE/start.d -nt /jetty-start ] ; then
+	if [ $(whoami) != "jetty" ]; then
+		cat >&2 <<- EOWARN
+			********************************************************************
+			WARNING: User is $(whoami)
+			         The user should be (re)set to 'jetty' in the Dockerfile
+			********************************************************************
+		EOWARN
+	fi
+
+	if [ -f $JETTY_BASE/jetty.start ] ; then
+		if [ $JETTY_BASE/start.d -nt $JETTY_BASE/jetty.start ] ; then
 			cat >&2 <<- 'EOWARN'
 			********************************************************************
 			WARNING: The $JETTY_BASE/start.d directory has been modified since
-			         the /jetty-start files was generated. Please either delete 
-			         the /jetty-start file or re-run /generate-jetty-start.sh 
+			         the $JETTY_BASE/jetty.start files was generated. Either delete 
+			         the $JETTY_BASE/jetty.start file or re-run 
+				     /generate-jetty.start.sh 
 			         from a Dockerfile
 			********************************************************************
 			EOWARN
 		fi
-		echo $(date +'%Y-%m-%d %H:%M:%S.000'):INFO:docker-entrypoint:jetty start command from /jetty-start
-		set -- $(cat /jetty-start)
+		echo $(date +'%Y-%m-%d %H:%M:%S.000'):INFO:docker-entrypoint:jetty start command from \$JETTY_BASE/jetty.start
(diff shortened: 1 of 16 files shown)


docker-entrypoint.sh
```diff
@@ -62,30 +62,40 @@ if expr "$*" : 'java .*/start\.jar.*$' >/dev/null ; then
...
-		if [ $JETTY_BASE/start.d -nt /jetty-start ] ; then
+	if [ $(whoami) != "jetty" ]; then
+		cat >&2 <<- EOWARN
+			********************************************************************
+			WARNING: User is $(whoami)
+			         The user should be (re)set to 'jetty' in the Dockerfile
+			********************************************************************
+		EOWARN
+	fi
+
```
md5, line 74: I think it would make sense to add the following here:

```shell
: ${JETTY_START:=$JETTY_BASE/jetty.start}
```

Then, all the places below that refer to `$JETTY_BASE/jetty.start` should be changed to `"$JETTY_START"`.

This allows a user to pass in `-e JETTY_START` to specify an alternate location that could be located on a volume or `tmpfs`. By doing so with this location and any others that need to be writable, they can run the container `--read-only`, as I mentioned at https://github.com/appropriate/docker-jetty/issues/80#issuecomment-334050251.

gregw: Good idea, but I'll set it the old fashioned way for `sh` in alpine.

docker-entrypoint.sh
```diff
@@ -62,30 +67,40 @@ if expr "$*" : 'java .*/start\.jar.*$' >/dev/null ; then
...
+	if [ -f $JETTY_START ] ; then
+		if [ $JETTY_BASE/start.d -nt $JETTY_START ] ; then
 			cat >&2 <<- 'EOWARN'
 			********************************************************************
 			WARNING: The $JETTY_BASE/start.d directory has been modified since
-			         the /jetty-start files was generated. Please either delete 
-			         the /jetty-start file or re-run /generate-jetty-start.sh 
+			         the $JETTY_START files was generated. Either delete
```
md5, line 80: Since we're using `'EOWARN'` above, this variable won't be interpolated. It seems like you want this to be interpolated. It's probably fine for `$JETTY_BASE` to be interpolated above as well, otherwise the `$` should be escaped.

md5 (approved):
This looks good to merge after the interpolation thing is sorted out.

I would probably double quote `$JETTY_START` most places it's being used in case someone uses a space in the path or something, but I also think that's pretty unlikely and can be left for a future enhancement if necessary.

docker-entrypoint.sh
md5: BTW, the syntax in my previous comment is POSIX syntax and works in Alpine's `/bin/sh` as well. See http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

md5: The colon is also a POSIX "special built-in": http://pubs.opengroup.org/onlinepubs/009695399/utilities/colon.html

docker-entrypoint.sh
gregw: I'm in two minds about expanding those variables, as they can make the error message verbose.  Reporting with the env var syntax can be more concise and comprehensible.

you call... I'm on the fence.

docker-entrypoint.sh
gregw: Oh really,  so I'm the really old fashioned one here.

Happy to update to the more concise syntax, but probably should do it for other variables

gregw: update to posix syntax

docker-entrypoint.sh
gregw: I tried the expansion and it reads OK because the paths are short, so committed

md5 (approved):
Ship it! :shipit:

PR #83, opened 2017-12-16
Script for inter container startup synchronization and container network verification 

janhelges:
### Ability to synchronize startup of jetty containers
This PR adds support to delay startup of a jetty container until a remote service become available. The feature is controlled by an environment variable as this is easy and is the preferred way to configure Docker-Compose / Kubernetes.

Typically useful when a database is running in a side container (a pod of jetty and a database) and the database container needs extra time to initialize before the database service become available to the jetty container.

The code in this PR is tested and is working in the standard jetty image using built-in bash supported and in an alpine jetty-container where the functionality is carried out by the netcat utility. 
 
### Usage example
Add the environment variable on the docker command line as in ...
 ```docker run -d --link databasehost:databasehost  -e WAIT_ON_HOST_PORT=databasehost:5432  ...```


gregw:
While this looks like a good feature, I think it is perhaps a bit too special purpose to be included in the main image. If you a start including features like these, there are always more features and some features age and eventually you end up with a big mess of env variables to maintain.

I'd much rather consider a mechanism that makes the entrypoint easily extensible by just adding scripts to a directory, so you could just add this script in a Dockerfile.   In the openjdk/jetty images for the  GoogleCloudPlatorm, the [entrypoint checks `/setupenv.d/*` for scripts to run ](https://github.com/GoogleCloudPlatform/openjdk-runtime/blob/master/openjdk-common/src/main/docker/docker-entrypoint.bash#L16-L21).  

@md5 what do you think of this approach to extensibility?

janhelges:
Hello @gregw  Thanks for quick response.
I can only say that for us it is a rather common usecase to  wait on a database service to become available before jetty shall go on and startup the war. I would also point out that this feature also may be used in an rather general way to achieve synchronization at container startup.

The alternatives to achieve this functionality is to ...
- write more or less the same script as in this PR to be run before the entrypoint script, and maybe exec the original entypoint script afterwards. This would be an fully alternative to the approach In the openjdk/jetty images for the GoogleCloudPlatorm you described earlier. The downside is that users have to write their own script and create their own jetty-image where they could achieved the same using plain environment configuration.
- not using jetty-dockerimage, but make an embedded jetty and take care of waiting in a startup script or in the java main. (Obviously it would be a poor alternative to roll-your-own instead of using the standard jetty container.)
- In kubernetes one may use the initcontainer-feature to control the synchronization between containers, but that solution is rather corse and will not be as snappy as one would like. 

I can understand your concern about many env variables over time. But still, I think this is a rather general case that justify an environment variable. Use of env-variables is very simple and **is** the preferred way to configure containers in Docker-Compose and Kubernetes. I regard this synchronization/waiting-issue as a rather common usecase in a multi container environment. I believe this PR adds a nice and rather general feature to the jetty dockerimage that would be beneficial to other users of a multi container environment where timing issues often could be a challenge.

gregw:
@janhelges, with my counter proposal, this functionality could be achieved simply by adding the script to a directory, so the entrypoint would not need to be either modified or replaced.

Potentially the script could even live in the default image and just be copied and/or linked to activate it.

janhelges:
I have reworked the PR based on feedback ...

Reverting to originally version: `/docker-entrypoint.sh`  and adding a new file with the synchronization support `/usr/local/bin/wait-before-docker-entrypoint`

### Usage...
```
docker run -itp8080:8080 --link dbhost:dbhost -e WAIT_ON_HOST_PORT=dbhost:5432  jetty:9.3-alpine   wait-before-docker-entrypoint
```
output...
```
Waiting for dbhost:5432 to become available
NC: dbhost:5432 is up
Now starting /docker-entrypoint.sh
2017-12-16 23:55:11.227:INFO::main: Logging initialized @226ms to org.eclipse.jetty.util.log.StdErrLog
2017-12-16 23:55:11.625:INFO:oejs.Server:main: jetty-9.4.7.v20170914
2017-12-16 23:55:11.664:INFO:oejdp.ScanningAppProvider:main: Deployment monitor [file:///var/lib/jetty/webapps/] at interval 1
...
```
This approach leaves the mainline unchanged as @gregw  suggested. At the same time it is easy to enable the feature just adding one command line argument together with an environment variable. 

This points out an easy way to extend the dockerimage with new functions/features in the future by adding scripts to `/usr/local/bin`

I hope this is better

gregw:
I like the approach a lot better (although it would be good to add documentation on usage to the comments of the script).
@md5 your thoughts?

janhelges:
I renamed the script to `network-probe`. I think this is a more appropriate name as it can be used alone or in combination with the /docker-entrypoint.sh or with other scripts on the jetty-dockerimage.

The `network-probe` utility may now be used with ...
- an environment-variable  or with ...
- command line arguments or with ...
- a combination of the two. 

The default behavior is to carry out the network probing and then execute the `/docker-entrypoint.sh` afterwards, but I added support to run this tool standalone using the `--exit` flag to allow docker-container-network troubleshooting using the jetty-dockerimage as a "debugging tool". The script now also support to hand over those arguments that is not interpreted by the script to the `/docker-entrypoint.sh`

I have also added documentation in the script. `network-probe --help` gives output...
```
Usage: ...
network-probe [arg1 [arg2 [arg3 ... [argN] ... ]]]

arg can be ...
 --help                     # Gives this help text and terminates
 --check=smtp.domain.com:25 # Reports if host smtp.mydomain.com port 25 is available
 --waitfor=dbhost:5432      # Wait until host dbhost port 5432 becomes available
 dbhost:1521                # Shorthand notation for --waitfor=dbhost:1521.
 --use-netcat-probe         # Uses the netcat probe instead of bash probe. Default is to use bash probe if bash is available.
 --waitfor-timestamp        # Include timestamp in --waitfor output, may be useful to see how long time it takes for a service to become available
 --no-waitfor-dot           # Do not print . in --waitfor output
 --exit                     # Exit. Terminate immediately without starting /docker-entrypoint.sh at the end as would be the default behavior

The args are processed in sequence.
For example ...
network-probe --waitfor=db:5432 --check=smtphost:25 --exit # this command will 1) wait for db:5432 and 2) check if smtphost:25 is up and 3) exit

It is also possible to set args in the environment variable NETWORK_PROBE_ARGS.
Args from the environment variable NETWORK_PROBE_ARGS is prepended to the command line args.
For example ...
export NETWORK_PROBE_ARGS="--check=smtp.domain.com:25 --waitfor-timestamp --waitfor=127.0.0.1:1521"
./network-probe oracledb:1521 # will 1) check smtp.domain.com:25 and 2) set timestamp in output and 3) wait for 127.0.0.1:5432, and 4) wait for oracledb:1521

Command line arguments that is not recognized by network-probe are handed over as command line arguments to /docker-entrypoint.sh
```

janhelges:
Hello @gregw  - hope you could help here. I get travis build errors on  `9.3-jre8` and `9.3-jre8-alpine` with lots of error messages like...
`gpg: keyserver receive failed: Address not available`
How could I fix the build for  `9.3-jre8` and `9.3-jre8-alpine` ?
Best regards Jan Helge

gregw:
@janhelges sorry I'm not familiar with the travis setup here.  Your branch builds cleanly for me. Waiting for comment from @md5  to decide what is best way forward

md5:
@janhelge @gregw The keyserver used for GPG keys doesn't differ between the images, so this appears to have been a transient failure. I've restarted the two failed builds.

As for whether this script belongs in the image, I would much rather see Jetty itself somehow signal that it is ready to accept connections and have the container use that to check it the container is ready to operate as opposed to probing the container over the network.

gregw:
@MD5, this script is waiting for a remote service to be available before starting jetty.  It is not waiting for jetty to be ready to accept connections.

We could potentially make it a jetty module that delays start until the remote server is available, but typically it is easier to write such behaviour in script rather than java.

I think my preference is still to have a `setup-env.d` directory into which derived images can drop arbitrary scripts that can setup, validate, waitfor the environment to be ready to proceed to start jetty.
I'm then neutral as to if this script should be in the image or somewhere else and linked to in documentation.

janhelge:
@gregw @md5 My experience is that timing issues and inter container networking challenges is typical problems in Docker compose and Kubernetes environment. The `network-probe` utility is a very simple tool, but still very useful in such environments.

With the default behaviour the `network-probe` chains nicely with the standard entrypoint, and it adds functionality to the Jetty image with very little "foot-print"

I'll be honored if you guys add the `network-probe` I you wish I be glad to contribute with more examples how to use this feature on the Jetty image documentation.

Best regards
Jan Helge

tianon:
I think it's probably also relevant to point out https://docs.docker.com/compose/startup-order/, which also links to https://github.com/vishnubob/wait-for-it, https://github.com/jwilder/dockerize, and https://github.com/Eficode/wait-for (which all seem conceptually similar to the proposed script).

gregw:
@tianon thanks for those links.  It would appear that this PR duplicates work that is already available.   The desired result can be achieved by adding one of these scripts to the image and using it as a wrapper to the existing entry-point.

So I'm going to close this PR.   @janhelge thanks for contributing the PR, even if ultimately it was not accepted due to existing solutions.

PR #89, opened 2018-05-21
Add "Architectures" to "generate-stackbrew-library.sh" output

### LINKED ISSUE #68: Review additional architecture support
As of https://github.com/docker-library/openjdk/pull/121, the `openjdk` base image supports additional architectures where the JRE is available on the given base OS. In practice, this means amd64, arm32v7, arm64v8, i386, ppc64le, and s390x for JRE 7 and 8, as well as amd64 for JRE 8 on Windows.

To the extent possible, I think we should have variants of the `jetty` image for these additional platforms and architectures.

tianon:
Closes #68

For repos under https://github.com/docker-library, we generally invoke `bashbrew` from within `generate-stackbrew-library.sh` to get the `Architectures` list of the image we're `FROM` directly, but I wasn't sure if you wanted to add that additional dependency for running the script here, so I wrote a quick one-liner to do the work as a one-off into files that get checked in (which also makes it easier to munge them if it turns out one of them fails to build or doesn't run properly etc and needs to be excluded).

As always, happy to adjust, rebase, amend, etc.  I could also set up Travis to test `i386`, if that's deemed useful (`memcached` has an example of that: https://github.com/docker-library/memcached/blob/5a09fc033d8e239ebecf1edb20b7f921d51de591/.travis.yml).

### DIFF
--- a/9.2-jre7/arches
+++ b/9.2-jre7/arches
@@ -1,0 +1,1 @@
+amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
--- a/9.2-jre8/arches
+++ b/9.2-jre8/arches
@@ -1,0 +1,1 @@
+amd64, arm32v5, arm32v7, arm64v8, i386, ppc64le, s390x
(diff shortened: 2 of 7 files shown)


tianon:
:tada: Want me to make the PR over to https://github.com/docker-library/official-images for you, or would you rather take that yourself?

md5:
That would be great!

PR #92, opened 2018-06-27
let users provide external volume for webapps in the Kitematic UI

buktt:
using the image with Kitematic - the UI that is shipped with docker - I cannot provide volumes.

The Kitematic uses the dockerfile to offer the user a UI for his volumes.
As long as the dockerfile does not declare it's volumes - the UI cannot do that...

### DIFF
--- a/9.4-jre8/Dockerfile
+++ b/9.4-jre8/Dockerfile
@@ -1,6 +1,6 @@
 FROM openjdk:8-jre
 
-# add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
+# add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added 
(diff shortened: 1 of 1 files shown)


buktt:
I don't understand, I'm new here - I'd appreciate somebody nice to help me in :)

The version I changed passed, and versions I did not touch failed?
What should I do to make them pass? I did not change anything in them...

gregw:
@buktt sorry this has not been attended to for a long time.
Can you tell me what the intent of your change is?  Why is the volume command needed rather than just use -v or --mount?

vutny (commented):
Looks like this has been dated and have conflicts.
I guess it could be closed, @gregw

gregw:
@buktt I have closed for now.  If you can explain why you needed a VOLUME command rather than just using -v or -mount, then we can consider another PR if needed.

PR #96, opened 2018-11-06
Allow runing as arbitrary user

mcanevet:
This is required for runing this image on OpenShift without having to
allow runing with RunAsAny SCC.


gregw:
@mcanevet sorry for slow response
Can you explain a bit more about why OpenShift can't have a group jetty and needs these files to be in root group?

mcanevet:
https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines
"By default, OpenShift Container Platform runs containers using an arbitrarily assigned user ID. This provides additional security against processes escaping the container due to a container engine vulnerability and thereby achieving escalated permissions on the host node.

For an image to support running as an arbitrary user, directories and files that may be written to by processes in the image should be owned by the root group and be read/writable by that group. Files to be executed should also have group execute permissions."

PR #97, opened 2018-11-12
update to 9.4.13.v20181111

gregw:
update to 9.4.13.v20181111

### DIFF
--- a/9.4-jre8/Dockerfile
+++ b/9.4-jre8/Dockerfile
@@ -8,7 +8,7 @@ ENV PATH $JETTY_HOME/bin:$PATH
 RUN mkdir -p "$JETTY_HOME"
 WORKDIR $JETTY_HOME
 
-ENV JETTY_VERSION 9.4.12.v20180830
+ENV JETTY_VERSION 9.4.13.v20181111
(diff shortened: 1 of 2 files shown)


md5 (approved):
I submitted a review this morning... Not sure why it didn't take. It looks like the CI failures are transient and should resolve with a retry.

gregw:
I'm going to hold off on this... as it appears that 9.4.13 may be a bad release and have a bad bug in it.    Thus hopefully we will have a 9.4.14 in a day or two, and I will then progress both at the same time (as I think we should have the 9.4.13 for completeness..  just not recommend it for use in most circumstances).

md5:
@gregw I'm not sure the best way to do that and still avoid having people download it as "9.4" during a small window. Also, I think there are other gaps in the Docker image version history for this image.

I would be inclined to skip the Docker image for 9.4.13.

@tianon thoughts?

tianon:
No opinion here -- if skipping 9.4.13 doesn't bother y'all, it's no problem for us. :+1:

gregw:
Let's skip 13... it's really living up to it's bad luck number and is a bit of a disaster of a release on several fronts!  14 a few days away.

Issue #98, opened 2018-12-21
Any plans for jre11 ?


s17t:
As the title. Are there plans for docker-jetty on jre 11 ?

gregw:
Latest jetty-9.4 releases are jre 11 ready, so I guess we should!  Hopefully soon...
We will also have a jetty-10 release early in the new year that is fully JPMSized

gregw:
I've created a PR #99 for this (along with upgrade to 9.4.14) and initial quick look seams OK.   @s17t could you give this branch a test?
I have also left out the alpine distribution for jre11 as I'm not sure if that makes sense anymore.

s17t:
Hello and thank you. I am testing right now and in the next days. I had to work around gid and uid of jetty user: in the alpine distro jetty user has 100:101 uid, gid but in debian it has 1000:1001. 

Nothing that can be fixed using usermod/groupmod and chown accordingly. I would state this somewhere in a changelog if the alpine tag will be dropped.

s17t:
> I'm not sure if that makes sense anymore.

It would made sense for sure since the image size is more than the double now. They are debating about alpine/openjdk11 issue in docker-library/openjdk/issues/211

md5:
#99 is closed, so Jetty 9.4 with JRE 11 will be available shortly. This is limited to Debian-based images, since Alpine-based images for JRE 11 don't exist yet.

PR #99, opened 2018-12-22
Issue #98 Upgrade to JRE-11 and Jetty-9.4.14

gregw:
Upgrade to JRE-11 and Jetty-9.4.14 for #98

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ install:
 
 before_script:
   - env | sort
+  - wget -qO- 'https://github.com/tianon/pgp-happy-eyeballs/raw/master/hack-my-builds.sh' | bash
   - cd "${VERSION}${VARIANT:+/$VARIANT}"
   - image="jetty:${VERSION}${VARIANT:+-$VARIANT}"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
   - VERSION=9.3-jre8 VARIANT=alpine
   - VERSION=9.4-jre8
   - VERSION=9.4-jre8 VARIANT=alpine
+  - VERSION=9.4-jre11
 
 install:
   - git clone https://github.com/docker-library/official-images.git ~/official-images
--- a/9.2-jre7/Dockerfile
+++ b/9.2-jre7/Dockerfile
@@ -35,7 +35,7 @@ RUN set -xe \
 	&& curl -SL "$JETTY_TGZ_URL.asc" -o jetty.tar.gz.asc \
 	&& export GNUPGHOME="$(mktemp -d)" \
 	&& for key in $JETTY_GPG_KEYS; do \
-		gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$key"; done \
+		gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys "$key"; done \
 	&& gpg --batch --verify jetty.tar.gz.asc jetty.tar.gz \
 	&& rm -rf "$GNUPGHOME" \
 	&& tar -xvf jetty.tar.gz --strip-components=1 \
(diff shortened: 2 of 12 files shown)


md5:
I think the `.travis-ci.yml` file needs to be updated as well: https://github.com/appropriate/docker-jetty/blob/master/.travis.yml#L6-L12

As far as the support matrix for Jetty and what version combinations make sense, I'll have to defer to you, but it seems weird that we're jumping from having JRE 8 as the maximum JRE version on all the Jetty releases to having a JRE 11 upgrade for only Jetty 9.4.

Also, I believe JRE 7 is EOL, so should that also be removed?

Lastly, is it possible to have a Jetty 9.4 + JRE 11 version on Alpine?

gregw:
I've updated `travis-ci.yml`

Java11 broke so many things that only the very latest jetty-9.4 is capable of running correctly on it.  We will never support 9.3 on JRE11.

As you say, JRE 7 and jetty 9.2 are both EOL, so we could remove them as there will be no more releases of either... unless there is a security problem that we wish to fix... so perhaps leave them a little while longer, but ultimately they can be dropped sooner or later.

There is not openjdk:11-jre-alpine to base a Jetty 9.4 + JRE 11 Alpine on, so I'm guessing that alpine is not appropriate for java11 or not yet available.

janblok:
Will this PR be merged?

tianon:
> There is not openjdk:11-jre-alpine to base a Jetty 9.4 + JRE 11 Alpine on, so I'm guessing that alpine is not appropriate for java11 or not yet available.

See https://github.com/docker-library/openjdk/issues/211#issuecomment-426708580 and the following discussion for why `openjdk:11-jre-alpine` doesn't exist (and isn't likely to).

janblok:
What is needed to get this PR merged?

gregw:
@janblok I'm investigating the CI failure and hopefully merged soon after that is fixed

gregw:
@md5 any idea why the gpg step is failing more often than not on CI?  Am I free to try other GPG keyservers?

md5:
@gregw This looks promising: https://github.com/inversepath/usbarmory-debian-base_image/issues/9#issuecomment-451635505

The "gpg: keyserver receive failed: Cannot assign requested address" error seems like it would jibe with an IPv6-related issue.

tianon:
See also https://github.com/docker-library/official-images/issues/4252,
especially https://github.com/docker-library/php/pull/666. 😅

gregw:
Thanks!
I'll try the no IPv6 first as it looks simpler than the happy eyeballs

gregw:
Hmmm I'm having no joy.   Need to install dirmngr, but that fails on alpine and on the others it doesn't improve the reliability of the gpg anyway!
This failure is not really related to this PR... @md5 can we merge without a clean CI?

md5:
> This failure is not really related to this PR... @md5 can we merge without a clean CI?

My concern about merging without clean CI is two-fold. The first is that this problem will likely happen on the official images build system as well, though perhaps @tianon has more insight there. The second is that any future PRs will probably also break with this same issue with the same frequency.

I think both of my concerns can be set aside to get this merged, assuming the image will build more reliably on the official images build system.

tianon:
Would it help if I put up a strawman PR to include pgp-happy-eyeballs into
CI here for easier review?

md5:
@tianon that would be great!

gregw:
I merged @tianon 's #101 PR here and we now build in CI, but I'm a bit cautious about downloading and running a script from a private repo! @md5 ?

md5:
@gregw I think you're right that the optics could be better, but in this case I think @tianon has as much control over @docker-library, so it is basically the same thing. @tianon is part of @infosiftr, who has partnered with @docker on the Official Images program pretty much from its inception as I understand it. @tianon is one of the primary maintainers of the `php` image mentioned above, the `debian` images that we rely on, and a bunch of other official images.

gregw:
@md5 well I'm happy if you are .... but as you're doing the trusting, I'll let you push the green buttons on both PRs :)

md5:
@gregw I'll open a `docker-library` PR

PR #101, opened 2019-02-05
Add "gpgconf --kill all" (and pgp-happy-eyeballs in CI) to help cut down on general GPG keyserver flakiness

tianon:
As discussed in https://github.com/appropriate/docker-jetty/pull/99#issuecomment-460510162.

This also adds `--batch` (see https://bugs.debian.org/913614 and PRs linked from docker-library/busybox#55).

### DIFF
--- a/9.2-jre7/Dockerfile
+++ b/9.2-jre7/Dockerfile
@@ -35,7 +35,7 @@ RUN set -xe \
 	&& curl -SL "$JETTY_TGZ_URL.asc" -o jetty.tar.gz.asc \
 	&& export GNUPGHOME="$(mktemp -d)" \
 	&& for key in $JETTY_GPG_KEYS; do \
-		gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$key"; done \
+		gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys "$key"; done \
 	&& gpg --batch --verify jetty.tar.gz.asc jetty.tar.gz \
 	&& rm -rf "$GNUPGHOME" \
 	&& tar -xvf jetty.tar.gz --strip-components=1 \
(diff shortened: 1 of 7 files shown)


gregw:
So this has passed the CI, which is a good change!
I still get some local failures with this, but building locally doesn't use the hacked `/etc/docker/daemon.json`, only the Dockerfile changes.

But I have also merged this with my #99 PR and that has now built!!!  However, I am a little cautious about merging in code that pulls a script from @tianon's personal repository and runs it on the CI machine.... but I'll defer to @md5 to approve/deny that.

9.2-jre7/Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN set -xe \
 	&& curl -SL "$JETTY_TGZ_URL.asc" -o jetty.tar.gz.asc \
 	&& export GNUPGHOME="$(mktemp -d)" \
 	&& for key in $JETTY_GPG_KEYS; do \
-		gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$key"; done \
+		gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys "$key"; done \
 	&& gpg --batch --verify jetty.tar.gz.asc jetty.tar.gz \
 	&& rm -rf "$GNUPGHOME" \
```
md5, line 40: Any reason not to add `gpgconf --kill all` to this `Dockerfile`? I suppose it is for JRE 7, so this Dockerfile should probably just be retired.

tianon: That'd be because this is based on Jessie which still uses GnuPG 1 so doesn't have `dirmngr` or `gpgconf` (nor daemons that get started automatically and need killing).

Issue #111, opened 2019-09-02
TMPDIR /tmp/jetty not used


pstrh:
It seems that the environment variable `TMPDIR /tmp/jetty` that is set up in the Dockerfile is not used. Instead the /tmp folder (default system property) is used.

```
docker run -it jetty --list-config
// prints: java.io.tmpdir = /tmp
```

From my understanding the dry-run command that generated the jetty.start configuration must be called with the -Djava.io.tmpdir JVM argument.

PR #112, opened 2019-09-05
Issue #111 TMPDIR used

gregw:
#111 

### DIFF
--- a/9.2-jre7/docker-entrypoint.sh
+++ b/9.2-jre7/docker-entrypoint.sh
@@ -59,7 +59,9 @@ if expr "$*" : 'java .*/start\.jar.*$' >/dev/null ; then
 			--version |\
 			-v )\
 			# It is a terminating command, so exec directly
-			exec "$@"
+			JAVA="$1"
+			shift
+			exec $JAVA $JAVA_OPTIONS "$@"
 		esac
 	done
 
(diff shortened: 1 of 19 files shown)


9.2-jre7/docker-entrypoint.sh
```diff
@@ -59,7 +59,9 @@ if expr "$*" : 'java .*/start\.jar.*$' >/dev/null ; then
 			--version |\
 			-v )\
 			# It is a terminating command, so exec directly
-			exec "$@"
+			JAVA="$1"
+			shift
+			exec $JAVA $JAVA_OPTIONS "$@"
```
md5, line 64: I feel like this script is getting pretty complicated and it would ideally be less so. One thought I have about this particular hunk is that it could make sense to extract a function that does this interpolation of `$JAVA_OPTIONS`. It also feels like that should happen right after the `# this is a command to run jetty` line, perhaps putting everything back into `$@` so that most of the script can still just `exec "$@"`?

md5 (approved):

9.2-jre7/docker-entrypoint.sh
md5: The separate function may not be necessary if the interpolation happens after that `# this is a command to run jetty` line, but I didn't attempt to make the changes myself

gregw: I kind of agree, but will merge like this while I ponder a better way to simplify

PR #118, opened 2019-11-19
Update supported architectures and base images

tianon:
This drops JRE 7, Alpine, and updates supported architectures to account for upstream changes (https://github.com/docker-library/openjdk/pull/322).

I believe this is what's got Travis failing over on https://github.com/docker-library/official-images/pull/6998.

If you want to continue supporting other architectures, I'd recommend checking out AdoptOpenJDK which does support more architectures (and is still very vanilla, although is Ubuntu-based instead of Debian-based, so may be more of a breaking change depending on how folks are using `jetty`).


gregw (changes requested):
It looks like the travis CI is still trying to build the jre7 and alpine versions.
So this PR should at  remove the alpine/jre7 references from .travis.yaml and from the Makefile all target.

gregw:
@md5 removing the alpine versions feels a bit drastic... is this what we should be doing?

md5:
I would defer to @tianon on this

tianon:
Doh sorry, I've updated `.travis.yml` now, but I left `Makefile` since it's using the make `wildcard` function which doesn't mind having an extra glob -- happy to remove that too though.

On the subject of Alpine, the OpenJDK project does not officially support Alpine/musl, which is why it was dropped from the `openjdk` image except for the EA builds, which then get dropped at GA (see Project Portola for more information about the upstream efforts to support musl properly).

If you want to continue to support Alpine-based images, your only options are `openjdk:14` (which will likely go away at 14's GA just like 13, 12, etc did) or `ibmjava` (which might see less and less maintenance now that IBM is pushing `adoptopenjdk` instead).

gregw:
I think an approach of pruning supported builds and then adding new ones by demand is best.

gregw (approved):

PR #122, opened 2019-11-27
Issue 113 keys multi build and 103 slim JDKs (with base image)

gregw:
This is a variant of #121 to solve #113 and #103 

It creates a base 9.4-jdk13 images, which all the other 9.4 variants are based off.


9.4-jdk13-slim/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# DO NOT EDIT. Edit Dockerfile-9.4 and use update.sh
+FROM openjdk:13-jdk-slim
+
+ENV JETTY_HOME /usr/local/jetty
+ENV JETTY_BASE /var/lib/jetty
+ENV TMPDIR /tmp/jetty
+ENV PATH $JETTY_HOME/bin:$PATH
+
+COPY --from=jetty:9.4-jdk13 $JETTY_HOME $JETTY_HOME
+COPY --from=jetty:9.4-jdk13 $JETTY_BASE $JETTY_BASE
```
vutny, line 10: I don't think that copying from external image is the best approach.
It means you should have it built already before doing other builds. Docker will not pull it automagically.
The issue is seen in Travis, it was unable to find non-existing image. And there is no way to build particular images in isolation.

I suggest to include all the commands to build basic Jetty 9.4 container (let's assume on `openjdk:latest`) as a first stage of all `Dockerfile`s for creating images with `9.4`. Since you're going to use templating with the script, it is rather easy to do.
This way `Dockerfile`s would became independent of each other, and layers cache would be used as well to speed up the process when building multiple images (in CI).

gregw: So you are saying that #121 is the best approach?

vutny: Seems to be so, I haven't looked too deep on it in the first place... Great minds think alike :smile:

PR #127, opened 2019-12-24
update to jetty 9.4.25

gregw:


vutny (changes requested):
Won't you wanna update root `Dockerfile-9.4` file as well?

vutny:
Looks like you need to update the version in both `/Dockerfile-9.4` and `/Dockerfile-9.4-slim`, @gregw .

gregw:
closing in favour of #128 which has a fix to the update.sh script to modify the Dockerfile-9.4 (and slim) files

PR #128, opened 2020-01-15
Fix arch, update and jdk13 

gregw:
Updates as per feedback from https://github.com/docker-library/official-images/pull/7134


vutny (commented):
Nice! Looking forward for this to be merged.

janbartel (approved):
LGTM

PR #132, opened 2020-01-27
Add support of Jetty9.4 on top of OpenJDK 8 JDK

PhantomYdn:


gregw (commented):
I'm wondering if we should have a policy that all slim builds are JRE and all other builds are JDK?
The idea is to keep the number of builds smaller.  Would this work for the OP?

PhantomYdn:
@gregw , I understand your idea, but in our case it will not help - we need JDK and also we are happy to have slim version. Might be it makes sense to use ARGs for dynamic original image? 
Example:
https://www.jeffgeerling.com/blog/2017/use-arg-dockerfile-dynamic-image-specification
So in sources you can have just few Dockerfiles - but in hub.docker.com specify as many builds - as needed.

PR #138, opened 2020-06-10
update readme to point to the new repository location

lachlan-roberts:


README.md
```diff
@@ -1,9 +1 @@
-# About this Repo
-
-[![Build Status](https://travis-ci.org/appropriate/docker-jetty.svg?branch=master)](https://travis-ci.org/appropriate/docker-jetty)
-
-This is the Git repo of the official Docker image for [jetty](https://registry.hub.docker.com/_/jetty/). See the
-Hub page for the full readme on how to use the Docker image and for information regarding contributing and issues.
-
-The full readme is generated over in [docker-library/docs](https://github.com/docker-library/docs),
-specificially in [docker-library/docs/jetty](https://github.com/docker-library/docs/tree/master/jetty).
+# Repository moved to https://github.com/eclipse/jetty.docker
```
gregw, line 1: I think this should be a bit better than a half sentence.

By all means lead with a big headline about how the repo has moved.    But then have a paragraph about what the repo is, how it was started by appropriate and then moved to co-reside with the jetty project at eclipse.

gregw (changes requested):

gregw (approved):

gregw:
@md5 what do you think?

gregw:
@md5 nudge

md5 (approved):
Looks fine. Thanks for the updates

md5:
That being said, I would also delete all the files other than the `README.md`. This repo should also be marked as archived, which I can do if necessary.

gregw:
@md5, I think we need to keep the files here until at least we have had the official images updated.  Otherwise people will follow those links and whilst the hex codes will still point to valid targets, it may be a little confusing.    So how about we merge this, then update official images and then come back here and clean up.

md5:
The official image docs have generated links back to the specific SHAs used
to build the tags, so I wouldn't worry about that

joakime:
> That being said, I would also delete all the files other than the `README.md`. This repo should also be marked as archived, which I can do if necessary.

Another option ...
You can create a new orphaned git branch which has no history, isn't based on any other branch, and only has a README.md

``` shell
cd repository
cp README.md /tmp/
git checkout --orphan archived_repo
git rm -rf .
rm '.gitignore'
cp /tmp/README.md .
git add README.md
git commit -a -m "Archived Readme"
git push origin archived_repo
```

Make that new orphaned git branch the default branch in this repo (default branch is a setting in github).
Then archive the repo in github.

An example of this would be the `master` branch on the https://github.com/eclipse/jetty.project/tree/master ("master" is not our default branch)
Compare that to the default branch view at https://github.com/eclipse/jetty.project

md5:
I'm not sure why we would want to disconnect the history; it seems useful
to have it readily available, not in a different branch.

The reason I think it's prudent to remove the files from HEAD is to avoid
having someone think they are current, not to render them less accessible.

joakime:
The history is still present and all of the existing shas still link.
There's no deletion, no removal.
It just a branch that is blank, isn't associated with any existing branch, or any existing tag, or any existing version, and used to community to users.

When you remove files from HEAD on the active default branch you are actually changing history and making it difficult for people to look things up, troubleshoot, follow history, or even fork and do their own thing.

Think of a new orphan branch as a "placeholder" or a "blank page" to get the end result you want.
No changing of history.
Just a README.md that shows in the github UI for people that go to the main page of the archived project.

All of the valid and useful branches still exist.
All of the tags still exist.
Even the state of the previously active default branch is still exists.

